PR #405, opened 2024-07-29
Adds writing to .env file from secrets in environment

alexrisch:


.github/workflows/build-internal.yml
```diff
@@ -29,6 +29,17 @@ jobs:
           eas-version: latest
           token: ${{ secrets.EXPO_TOKEN }}
 
+      - name: Create .env file from secrets
+        run: |
+          echo "EXPO_PUBLIC_CONTACT_ADDRESS=${{ secrets.EXPO_PUBLIC_CONTACT_ADDRESS }}" >> .env
```
nmalzieu, line 34: the way I set it up right now, it needs the file to be called `.env.production` for prod, and `.env` for preview. That makes it easier to have both locally, but if you want to change that we can

nmalzieu (changes requested):

.github/workflows/build-internal.yml
alexrisch: That's fine, I'll make the update here

nmalzieu (approved):

PR #409, opened 2024-07-29
feat: Reactions Improvements 

alexrisch:
Reactions Improvements with Context Menus

### DIFF
--- a/components/Chat/Message/Message.tsx
+++ b/components/Chat/Message/Message.tsx
@@ -1,3 +1,4 @@
+import { MessageContextMenuWrapper } from "@components/MessageContextMenuWrappers/MessageContextMenuWrapper";
 import {
   inversePrimaryColor,
   messageInnerBubbleColor,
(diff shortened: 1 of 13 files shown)


saulmc:
Got this up and running in the simulator and noted a few things:
- Slight white border around the highlighted message
- No message tail on the highlighted message
- Tap targets are misaligned -- tapping the right half of an emoji will highlight the next emoji
- Tapping an emoji does not actually add a reaction
- Button row placement, padding, and background color are different from designs

https://github.com/user-attachments/assets/000b89be-48b7-4433-995e-93270d8e929d

nplasterer (approved):

saulmc:
Really close on this! Just a few last items (see video):

https://github.com/user-attachments/assets/aeaf4902-fb80-4e19-b0be-44fdfa0b424c

- White rectangle in bubble after selecting emoji
- Some "ghosting" of reactions after closing the context menu

There are also a few follow-ups that I don't consider blockers for this PR:
- Dates are missing from above the message bubbles
- The + button in the design should expand the emoji picker, not add a + emoji. My apologies that this was not clearly specified and clearly the design is ambiguous. Let's use an icon (not emoji) and add background like Messenger does here:

IMG_4844

PR #424, opened 2024-07-30
feat: Added error display when client loading fails

alexrisch:
Added error display for client errors
Show text to user on profile

Simulator Screenshot - iPhone 15 Pro - 2024-07-30 at 07 56 45
Simulator Screenshot - iPhone 15 Pro - 2024-07-30 at 07 56 43
Simulator Screenshot - iPhone 15 Pro - 2024-07-30 at 07 56 41
Simulator Screenshot - iPhone 15 Pro - 2024-07-30 at 07 56 32
Simulator Screenshot - iPhone 15 Pro - 2024-07-30 at 07 56 13
Simulator Screenshot - iPhone 15 Pro - 2024-07-30 at 07 56 09

### DIFF
--- a/screens/Navigation/ConversationListNav.ios.tsx
+++ b/screens/Navigation/ConversationListNav.ios.tsx
@@ -1,3 +1,5 @@
+import { ErroredHeader } from "@components/ErroredHeader";
+import { useShouldShowErrored } from "@hooks/useShouldShowErrored";
 import { NativeStackScreenProps } from "@react-navigation/native-stack";
 import { textPrimaryColor } from "@styles/colors";
 import React, { useLayoutEffect } from "react";
@@ -8,6 +10,7 @@ import {
   View,
   useColorScheme,
   TouchableOpacity,
+  StyleSheet,
 } from "react-native";
 import { SearchBarCommands } from "react-native-screens";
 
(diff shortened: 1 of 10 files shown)


nmalzieu (approved):

screens/Navigation/ConversationListNav.ios.tsx
```diff
@@ -118,18 +114,22 @@ export default function ConversationListNav() {
...
+                      color:
+                        colorScheme === "dark"
+                          ? "rgba(255, 255, 255, 0.6)"
+                          : "rgba(0, 0, 0, 0.6)",
+                    },
+                  ]}
+                >
+                  {name}
+                </Text>
+                {shouldShowError && <ErroredHeader />}
```
saulmc, line 131: Could this replace the loading indicator, rather than be placed next to it?

utils/i18n/strings.ts
```diff
@@ -41,6 +41,7 @@ export const strings = {
   are_you_sure: "Are you sure?",
   promote_to_admin: "Promote to admin",
   send_a_message: "Send a message",
+  client_error: "Looks like something went wrong, please try reconnecting.",
```
saulmc, line 44: If I'm reading useShouldShowErrored correctly, this is actually a network state indicator rather than a generic error state. I think either the copy and presentation should reflect that, or, the error should react more directly to a client error rather than whether the internet is reachable.

IMG_4678
IMG_4679

alexrisch: @saulmc _should_ be the opposite https://github.com/ephemeraHQ/converse-app/pull/424/files#diff-0019535d4794401d7721f32799550bf423ef51d9630bfe9e4feb6fd2050b3200R8
Should only show if you are connected to the internet and the client creation failed

saulmc: Misread! Carry on. Still think it should replace the spinner since "never going to load" and "taking really long to load" are mutually exclusive.

screens/Navigation/ConversationListNav.ios.tsx
saulmc: 👍 Didn't realize the app can recover if the client creation failed.

PR #442, opened 2024-07-31
Noe/fix more crashes

nmalzieu:
A few fixes for crashes
- iOS crashes happening because of lock files that are still there in specific cases
- logout crashes because we try to access non existent zustand store because it has been deleted
- dropping connection for the Converse one because we stop hitting it but a READ lock is still present (which should not be an issue in WAL unencrypted mode but still let's make sure we kill the connection)
- wait for user interaction before launching the "authenticate to inbox" signature => should prevent some signature loops!

Also, some cleanup in logs & DebugButton + syncing groups when listing them


cameronvoell (approved):
looks good, spot testing dev build looked good as well 👌

PR #452, opened 2024-08-02
Noe/export logs

nmalzieu:
Implement a logging system for Converse

- logging to terminal in debug mode
- writing to file 
- adding sentry breadcrumbs or logging a sentry exception depending on severity
- replaced all console.log with converse logger
- moved the debug menu on shake
- added a way to export the .log file from the shake and to reset the log session (i.e. rotate log file)
- enabled debug mode for everyone in prod

NB: before releasing to the outside world, we should

- disable debug mode for everyone in prod (only for a list of ephemera addresses for instance)
- disable file logging when not in debug mode (might impact perf)
- find a way to show debug menu during onboarding (because no address yet)

@saulmc should we open an issue to make sure we fix this before releasing?


alexrisch (approved):

cameronvoell:
> NB: before releasing to the outside world, we should
> disable debug mode for everyone in prod (only for a list of ephemera addresses for instance)
> disable file logging when not in debug mode (might impact perf)
> find a way to show debug menu during onboarding (because no address yet)

Created an issue here: https://github.com/ephemeraHQ/converse-app/issues/455 for disabling in prod before release

utils/shake.ts
```diff
@@ -0,0 +1,36 @@
+import { Accelerometer, AccelerometerMeasurement } from "expo-sensors";
+import { useState, useEffect } from "react";
+
+const THRESHOLD = 4; // Adjust this value based on your sensitivity preference
```
alexrisch, line 4: In testing on Android this feels pretty hard to get the dev menu to open also I'm seeing some pretty bad Android performance on this branch comparing to main

PR #470, opened 2024-08-07
Noe/crash on load

nmalzieu:
Seems to fix the crash on wallet click (version 237 on testflight works for me)
Also fixes local building that was failing on my computer

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@xmtp/content-type-transaction-reference": "^1.0.3",
     "@xmtp/frames-client": "^0.5.0",
     "@xmtp/proto": "^3.60.0",
-    "@xmtp/react-native-sdk": "^2.2.3",
+    "@xmtp/react-native-sdk": "^2.2.4",
     "@xmtp/xmtp-js": "11.5.0",
     "amazon-cognito-identity-js": "^6.3.12",
     "axios": "^1.2.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -23,8 +23,14 @@
   },
   "dependencies": {
     "@azure/core-asynciterator-polyfill": "^1.0.2",
+    "@babel/core": "^7.25.2",
     "@babel/plugin-proposal-decorators": "^7.20.7",
     "@babel/plugin-proposal-export-namespace-from": "^7.18.9",
+    "@babel/plugin-transform-class-properties": "^7.24.7",
```
alexrisch, line 29: I don't think babel should be required in dependencies rather than devDependencies

Is there a reason these are being moved?

alexrisch:
Rebased with latest main

package.json
nmalzieu: Yes building was failing due to missing dependencies, I'm not sure why, and followed that comment
https://github.com/facebook/create-react-app/issues/13189#issuecomment-1686243701

nmalzieu: I agree this is not ideal but wonder if there is any risk?

alexrisch: Bundle size would be the biggest concern, and if this workaround is masking a different issue

alexrisch (approved):

alexrisch:
Made an issue to track in the future, but would like to include these changes in the next build

PR #497, opened 2024-08-10
feat: place reactions outside bubbles for attachments, frames, transactions

saulmc:
- Places reactions outside of bubbles for attachment, frame, and transaction messages
- Removes inconsistent borders, absolute positioning from all messages

### DIFF
--- a/components/Chat/Message/Message.tsx
+++ b/components/Chat/Message/Message.tsx
@@ -583,8 +583,7 @@ const useStyles = () => {
       marginBottom: 8,
     },
     outsideReactionsContainer: {
-      flexShrink: 1,
-      flexGrow: 1,
+      flex: 1,
     },
   });
 };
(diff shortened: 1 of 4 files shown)


components/Chat/Message/Message.tsx
```diff
@@ -571,5 +571,20 @@ const useStyles = () => {
...
+      columnGap: 8,
+      width: "100%",
+    },
+    reactionsContainer: {
+      marginHorizontal: 8,
+      marginBottom: 8,
+    },
+    outsideReactionsContainer: {
+      flexShrink: 1,
+      flexGrow: 1,
```
alexrisch, line 587: This can be done as flex: 1

alexrisch, line 587:
```suggestion
      flex: 1,
```

alexrisch (approved):

PR #513, opened 2024-08-15
Performance Improvements

alexrisch:
Removed useClient

Moved entify into query
Updated QueryKeys
Made GroupAvatar not request on conversation list screen


queries/useGroupMessages.ts
```diff
@@ -13,16 +13,8 @@ export const useGroupMessages = (account: string, topic: string) => {
         return;
```
nmalzieu, line 13: shouldn't we return 
```
{
  byId: {},
  ids: []
}
```

here?

components/ConversationList/GroupConversationItem.tsx
```diff
@@ -38,9 +38,13 @@ export const GroupConversationItem: FC<GroupConversationItemProps> = ({
...
     refetchOnMount: false,
+    staleTime: Infinity,
+    gcTime: Infinity,
   });
   const { data: groupPhoto } = useGroupPhotoQuery(userAddress, topic, {
     refetchOnMount: false,
+    staleTime: Infinity,
+    gcTime: Infinity,
   });
   const { blockGroup } = useGroupConsent(topic);
```
nmalzieu, line 49: `useGroupConsent` will call many other queries and needs to be prevented to query the bridge also?

nmalzieu (changes requested):
Thanks, looks cleaner!

Just did a quick code review without testing as I think there's a bit more to do

Once it's done I'll give it a go, adding logs when querying the bridge, to see the improvements and if there are other places that need work!

components/ConversationList/GroupConversationItem.tsx
alexrisch: Good catch! I'll update for that too

alexrisch: Updated!
Verified it's not being called on the conversation list after initial load any more

nmalzieu:
I have successfully managed to break the app again on a big account with many groups.

By commenting some queries I can unblock it so I still think it's linked to react queries… I'm still unsure why it just locks the app, it's like the JS thread is blocked.

Small investigation:

- `useGroupNameQuery` used in the group list item calls `useGroupQuery` which calls `useGroupsQuery` : we should pass along queryOptions or it might trigger calls 
- same for `useGroupPhotoQuery` 
- `useGroupConsent` calls `useGroupQuery` but also `useGroupCreator` which also calls `useGroupQuery` they should probably also get the change

alexrisch:
Closing this out in favor of the other branch

PR #537, opened 2024-08-21
Tentative fix for db corruption

nmalzieu:
We saw cases where only group convos were showing
I suspect this is some db corruption introduced with the recent db disconnect feature because we save all group convos to db but only the 1:1 that are "new" so it would end up having only groups if there is a corruption

Tentative fix
- make sure we reconnect database even if hydration not done
- wait until app is back active before doing some critical database actions
- recover from a corruption by repulling all from network

### DIFF
--- a/utils/appState.ts
+++ b/utils/appState.ts
@@ -1,0 +1,28 @@
+import { AppState } from "react-native";
+
+import logger from "./logger";
+
+/*
+Method to wait until the app is in Foreground (Active)
+helpful to make sure we're not doing critical database stuff
+while the database is disconnected to prevent data loss
+Will check every delayMs, then wait endDelayMs for db reconnection
+*/
+export const waitUntilAppActive = async (
+  delayMs: number,
+  endDelayMs: number
+) => {
+  let appWasInactive = false;
+  while (AppState.currentState !== "active") {
+    appWasInactive = true;
+    logger.debug(`App is inactive, waiting ${delayMs}ms until back active`);
+    await new Promise((r) => setTimeout(r, delayMs));
+  }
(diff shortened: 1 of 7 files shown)


utils/appState.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import logger from "./logger";
+
+/*
+Method to wait until the app is in Foreground (Active)
+helpful to make sure we're not doing critical database stuff
+while the database is disconnected to prevent data loss
+Will check every delayMs, then wait endDelayMs for db reconnection
+*/
+export const waitUntilAppActive = async (
```
alexrisch, line 11: Approved, but I think we should avoid patterns like this especially when there are event listeners that handles this

`export const waitUntilAppActive = async (delayMs, endDelayMs, retries = 3) => {
  return new Promise((resolve, reject) => {
    let attempts = 0;

    const handleAppStateChange = (nextAppState) => {
      if (nextAppState === 'active') {
        logger.debug(`App is active, waiting ${endDelayMs}ms more`);
        setTimeout(() => {
          AppState.removeEventListener('change', handleAppStateChange);
          resolve();
        }, endDelayMs);
      } else if (attempts >= retries) {
        AppState.removeEventListener('change', handleAppStateChange);
        reject(new Error('App did not become active within the allowed retries.'));
      } else {
        attempts++;
      }
    };

    AppState.addEventListener('change', handleAppStateChange);
  });
};`

nmalzieu: @alexrisch apparently the PR is still unapproved?
I added a commit to use event handler, tell me if that works for you!

alexrisch (approved):

Issue #543, opened 2024-08-21
Feature request: Add ability to revoke Installations


alexrisch:

nmalzieu:
@saulmc product question

- to revoke installations for an inbox, you need to sign with the wallet that owns that inbox
- currently Converse can only be linked to one wallet : meaning if you want to revoke installations for inbox A - then for inbox B, you have to connect with different wallets. Is that ok for now (seems pretty edge)

Also, we'll want at some point a place where you can
- see the current wallet app / wallet id that is currently linked to Converse
- disconnect
- connect a new one (right now the connect drawer appears only when you click on "Revoke other installations"

That work can be part of that PR or the transactional frames one that I could start right after, what do you think?

PR #544, opened 2024-08-21
feat: Added syncAllGroups

alexrisch:
Added syncAllGroups call
Closes https://github.com/ephemeraHQ/converse-app/issues/543

### DIFF
--- a/i18n/translations/en.ts
+++ b/i18n/translations/en.ts
@@ -58,7 +58,7 @@ const en = {
   revoke_other_installations: "Revoke other installations",
   revoke_all_other_installations: "Revoke all other installations",
   revoke_description:
-    "Revoke all other group installations of XMTP on other applications and devices.",
+    "Revoke all other V3 installations of XMTP on other applications and devices.",
 
   // Attachments
   photo_library: "Photo Library",
--- a/screens/Profile.tsx
+++ b/screens/Profile.tsx
@@ -1,6 +1,6 @@
 import Picto from "@components/Picto/Picto";
 import { useShouldShowErrored } from "@hooks/useShouldShowErrored";
-import { translate } from "@i18n/translate";
+import { translate } from "@i18n";
 import Clipboard from "@react-native-clipboard/clipboard";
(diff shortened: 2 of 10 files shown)


i18n/translations/en.ts
```diff
@@ -46,6 +46,17 @@ const en = {
...
+  common_activity: "COMMON ACTIVITY",
+  social: "SOCIAL",
+  address: "ADDRESS",
+  youre_the_og: "YOU'RE THE OG",
+  app_version: "APP VERSION",
+  security: "SECURITY",
+  revoke_other_installations: "Revoke other installations",
+  revoke_all_other_installations: "Revoke all other installations",
+  revoke_description:
+    "Revoke all other group installations of XMTP on other applications and devices.",
```
nplasterer, line 59:
```suggestion
    "Revoke all other V3 installations of XMTP on other applications and devices.",
```

screens/Profile.tsx
```diff
@@ -330,6 +331,38 @@ export default function ProfileScreen({
     );
   }, [colorScheme, logout]);
 
+  const showRevokeActionSheet = useCallback(async () => {
```
nplasterer, line 334: Totally doesn't have to happen in this PR but might make sense to show the number of other installations. Like a `revoke 10 other installations`.

nplasterer: This can be retrieved from the new `InboxState`

utils/xmtpRN/conversations.ts
```diff
@@ -296,6 +296,7 @@ const listConversations = async (client: ConverseXmtpClientType) => {
 const listGroups = async (client: ConverseXmtpClientType) => {
   // @todo => this will be adapted once we have a syncAllGroups method
   const groups = await fetchGroupsQuery(client.address);
+  // client.conversations.syncAllGroups();
```
nmalzieu, line 299: goal of using this line is to remove the loop right after

nmalzieu (changes requested):

nplasterer (approved):

PR #547, opened 2024-08-23
Revoke other installations when logging in (temporary feature)

nmalzieu:
- When logging in, we want to revoke other installations because many of them are broken
- Also implements a check to logout user from other devices if installation has been revoked
- + check during onboarding if logging in with an existing libxmtp database but installation has been revoked

N.B. : due to the fact that we disconnect the XMTP db when leaving the app, I hit issues with `Client error: storage error: Pool needs to  reconnect before use` when coming back to the app after the revoke signature. 

Since this is a temporary debug feature for now, I implemented an ugly fix (reconnecting the database in this specific case when we leave the app). We can do that since during the onboading the db is not in the shared container.

But we will hit this kind of issues in the future:
- when revoking installations from the settings page (not the onboarding)
- when adding new EOA / SCW
- any feature that requires a signature and has an impact in the libxmtp database

I think we need a way to "wait and retry" when we hit a `Client error: storage error: Pool needs to  reconnect before use`
This will probably need to be done in libxmtp?

Also: revoked devices will need to "logout & delete libxmtp database", if not they will re-login next time with the same installation key. Should we implement a "logout & delete installation" while keeping group history?

cc @neekolas @nplasterer

### DIFF
--- a/components/DebugButton.tsx
+++ b/components/DebugButton.tsx
@@ -7,6 +7,8 @@ import {
 } from "@utils/logger";
 import { navigate } from "@utils/navigation";
 import { getNativeLogFile } from "@utils/xmtpRN/logs";
+import { getXmtpClient } from "@utils/xmtpRN/sync";
+import { Client } from "@xmtp/react-native-sdk";
 import axios from "axios";
 import Constants from "expo-constants";
 import { Image } from "expo-image";
@@ -80,6 +82,18 @@ const DebugButton = forwardRef((props, ref) => {
             return Alert.alert("No previous session logging file found");
           }
           navigate("WebviewPreview", { uri: previousLoggingFile });
+        },
+        "Get installations": async () => {
+          const client = (await getXmtpClient(currentAccount())) as Client;
+          const state = await client.inboxState(true);
+          Alert.alert(
+            `${state.installationIds.length} installations`,
+            `InboxId: ${client.inboxId}\n\nCurrent installation: ${
+              client.installationId
+            }\n\nAll installations:\n\n${state.installationIds
+              .map((i) => `${i}`)
+              .join("\n\n")}`
+          );
         },
         "Trigger OTA Update": async () => {
           try {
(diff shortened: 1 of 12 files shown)


cameronvoell:
hey @nmalzieu  - know this is a temporary fix. My understanding is that this build will not work on more than one device simultaneously, does that sound right to you? I think it's fine since this is temporary, but just wanted to make sure we call that out loudly. 

> I think we need a way to "wait and retry" when we hit a Client error: storage error: Pool needs to  reconnect before use
This will probably need to be done in libxmtp?

I can look into whether this is possible. I'm thinking it will take some coordination between libxmtp and the SDKs. 

> Should we implement a "logout & delete installation" while keeping group history

I think this may be possible, need to think through a bit more if this would be useful outside this temporary situation we're in now, with some inboxes with invalid installation keys.

nmalzieu:
> hey @nmalzieu - know this is a temporary fix. My understanding is that this build will not work on more than one device simultaneously, does that sound right to you? I think it's fine since this is temporary, but just wanted to make sure we call that out loudly.
> 
> > I think we need a way to "wait and retry" when we hit a Client error: storage error: Pool needs to  reconnect before use
> > This will probably need to be done in libxmtp?
> 
> I can look into whether this is possible. I'm thinking it will take some coordination between libxmtp and the SDKs.
> 
> > Should we implement a "logout & delete installation" while keeping group history
> 
> I think this may be possible, need to think through a bit more if this would be useful outside this temporary situation we're in now, with some inboxes with invalid installation keys.

@cameronvoell yes indeed currently this will work only on one device at a time. Let's confirm but I think @neekolas was OK with that.

cameronvoell:
@nmalzieu any specific reason you set this back to draft?

nmalzieu:
> @nmalzieu any specific reason you set this back to draft?

as discussed during standup I needed to make it possible to login without revoking so people can still use multiple device. We will communicate with the team to do it at least once. Now ready for review!

cameronvoell (approved):
tried the flow, worked like a charm 👌

PR #553, opened 2024-08-23
fix: Android Notification Linking

alexrisch:
Fixed notification handling for Android
Fixes https://github.com/ephemeraHQ/converse-app/issues/373
Fixes https://github.com/ephemeraHQ/converse-app/issues/552


nmalzieu (approved):
Approved!
So ios has data parsed in `event.notification.request.content.data` and android has it as a string in `event.notification.request.content.data.body` ? Account switching still works on iOS right? Weird!

PR #582, opened 2024-08-27
fix: sharing files from debug screen fixed so it works on iOS and Android

cameronvoell:
Closes https://github.com/xmtp/xmtp-react-native/issues/483

### DIFF
--- a/ios/Gemfile.lock
+++ b/ios/Gemfile.lock
@@ -1,26 +1,29 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    CFPropertyList (3.0.6)
+    CFPropertyList (3.0.7)
+      base64
+      nkf
       rexml
     activesupport (7.0.7.2)
       concurrent-ruby (~> 1.0, >= 1.0.2)
       i18n (>= 1.6, < 2)
       minitest (>= 5.1)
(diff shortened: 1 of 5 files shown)


ios/Gemfile.lock
```diff
@@ -1,26 +1,29 @@
...
+    addressable (2.8.6)
       public_suffix (>= 2.0.2, < 6.0)
     algoliasearch (1.27.5)
       httpclient (~> 2.8, >= 2.8.3)
       json (>= 1.5.1)
     atomos (0.1.3)
+    base64 (0.2.0)
     claide (1.1.0)
-    cocoapods (1.12.1)
+    cocoapods (1.15.2)
```
cameronvoell, line 21: this Gemfile.lock file was forcing my CLI to use cocoapods 1.12.1 which would make both `npx pod-install` and `pod install` command break for me bc 1.12.1 is not compatible with our project (missing a bunch of visionOS dependencies).

In any case, I can tell others also arent using cocoapods 1.12.1 because the Podfile.lock shows cocoapods 1.15.2, so  I think this Gemfile.lock can be updated.

alexrisch (approved):

Issue #584, opened 2024-08-28
Bug: Group Invites Show as Pending when they were actually added to the group


alexrisch:
### Describe the bug

Looks like putGroupInviteRequest in ios/ConverseNotificationExtension/NotificationService.swift isn't successful and need to find out why

PR #586, opened 2024-08-28
fix: Don't show pending requests for users already in the group

alexrisch:
Filter out requests that have already been approved but not updated on the backend Set storage so users can only attempt to join once from app installation perspective

Closes https://github.com/ephemeraHQ/converse-app/issues/584 
Closes https://github.com/ephemeraHQ/converse-app/issues/588

### DIFF
--- a/hooks/useGroupPendingRequests.ts
+++ b/hooks/useGroupPendingRequests.ts
@@ -3,20 +3,25 @@ import { PendingGroupJoinRequest } from "@utils/api";
 import { useMemo } from "react";
 
 import { useExistingGroupInviteLink } from "./useExistingGroupInviteLink";
+import { useGroupMembers } from "./useGroupMembers";
 
 export const useGroupPendingRequests = (topic: string) => {
   const { data } = usePendingRequestsQuery();
+  const { members } = useGroupMembers(topic);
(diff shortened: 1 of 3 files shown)


neekolas:
Curious why requests would be approved but not updated on the backend. Shouldn't the two things happen at roughly the same time?

alexrisch:
> Curious why requests would be approved but not updated on the backend. Shouldn't the two things happen at roughly the same time?

They should, and from debugging it looks like it's being called and works successfully locally from what I am seeing
I think what may happen is a second request gets made, but I am not 100% sure

Either way it should clean up some UX

neekolas:
Just feels a bit like we are covering up the issue rather than resolving it. 

If there are duplicate requests shouldn't we either prevent them from being sent in the first place or respond to them?

We can ticket this to handle later if we think that's going to be a pain.

alexrisch:
@neekolas this PR adds the part of caching the invite https://github.com/ephemeraHQ/converse-app/pull/586/files#diff-8f2c7accf2c2da40406bf00e47a5197cde61da72f12d858f33df6ccdf56323c8R69

nmalzieu (approved):

PR #616, opened 2024-08-29
Remove .git from repositoryURL

nmalzieu:
When switching between CLI `pod install` and Xcode builds, an annoying `.git` gets added/removed to the package resolution.
Fixing it by removing `.git` from the `repositoryURL`. Verified it still builds 2 ways.

### DIFF
--- a/ios/Converse.xcodeproj/project.pbxproj
+++ b/ios/Converse.xcodeproj/project.pbxproj
@@ -321,7 +321,7 @@
 			);
 			mainGroup = 83CBB9F61A601CBA00E9B192;
 			packageReferences = (
-				4C31F4EC29AF887B0032D062 /* XCRemoteSwiftPackageReference "Alamofire.git" */,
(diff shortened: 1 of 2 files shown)


ios/Converse.xcodeproj/project.pbxproj
```diff
@@ -1013,9 +1013,9 @@
 /* End XCConfigurationList section */
 
 /* Begin XCRemoteSwiftPackageReference section */
-		4C31F4EC29AF887B0032D062 /* XCRemoteSwiftPackageReference "Alamofire.git" */ = {
+		4C31F4EC29AF887B0032D062 /* XCRemoteSwiftPackageReference "Alamofire" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/Alamofire/Alamofire.git";
+			repositoryURL = "https://github.com/Alamofire/Alamofire";
```
nmalzieu, line 1018: this is the one that fixes the issue

alexrisch:
Sorry I think it got added when I was on an older code version

PR #617, opened 2024-08-29
Design touchup for quick reactions

saulmc:
Aligns quick reactions menu with designs. Hides reaction attribution component, which needs to be affixed to the screen top.

### DIFF
--- a/components/Chat/Message/MessageReactionsList.tsx
+++ b/components/Chat/Message/MessageReactionsList.tsx
@@ -7,9 +7,9 @@ import {
 import { useSelect } from "@data/store/storeHelpers";
 import {
   backgroundColor,
-  clickedItemBackgroundColor,
   textPrimaryColor,
   textSecondaryColor,
+  tertiaryBackgroundColor,
 } from "@styles/colors";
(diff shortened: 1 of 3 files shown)


components/Chat/Message/MessageReactionsList.tsx
```diff
@@ -230,7 +231,7 @@ const MessageReactionsListInner: FC<MessageReactionsListProps> = ({
         message.fromMe ? styles.fromMeContainer : styles.fromOtherContainer,
       ]}
     >
-      {list.length !== 0 ? (
+      {/* {list.length !== 0 ? (
```
alexrisch, line 234: We want to remove this now?
The library has a pretty tough time handling it for sure.
If this is a hard requirement might need to move to fully JS instead, the problem I had with that is I wasn't sure what Android and web should look like

saulmc: My view is that this element isn't essential since there are also profile pics in the actual message bubble reactions. We can punt it and revisit the implementation approach and/or design. Most designs can be substituted with less expensive alternatives, especially if it's for a common messaging feature.

alexrisch (approved):

Issue #583, opened 2024-08-28
Feature request: Move Profile Fetching to React Query


alexrisch:
### Is your feature request related to a problem?

We have several issues related to profile resolution: https://github.com/ephemeraHQ/converse-app/labels/Name%20Resolution

We can look to remove profiles from the Converse App SQL Database and allow the React Query library to handle fetching and persistance

We can also use this to update our MMKV store for native usage with Notifications

### Describe the solution to the problem

Would like to try something with batching:
https://tanstack.com/query/v4/docs/framework/react/community/batching-requests-using-bathshit

The profiles query hook should be something like:

```
useQuery({
....
gcTime: 1000 * 60 * 60 * 24 * 30,
staleTime: 1000 * 60 * 60 * 24,
refetchOnMount: false,
})
```

PR #633, opened 2024-08-30
Profiles to React Queries

alexrisch:
Why?
Right now zustand is being used to manage the network state of profiles We refresh profiles whenever a new group is added/new conversation/group is refreshed

Theres some potential issues:
Groups can get large
Conversation lists can get large
New and interesting ways of using profiles in the app may be added (group invites for example) These require adding additional checks if we should update the profiles etc. What happens when a request fails? What happens if members change? As we move towards open source it can be helpful to devs to have more consumable hooks, rather than tracking down exactly where the data is coming from and why it's refreshing/cached/missing

With React Query we get better out of the box retries, and can use a batching and deduplicating library to manage this for us The main goal of this is that incase anything gets missed, we can have a foolproof fallback to refresh just in case

We can further improve the usage of profiles by:
1. Having MMKV entries for each profile rather than the whole state
2. Remove fetching in group syncs and have a full fetch for each groups members later on
3. Completely removing the Zustand profile store to avoid storing network state in Zustand

Closes https://github.com/ephemeraHQ/converse-app/issues/583

### DIFF
--- a/queries/QueryKeys.ts
+++ b/queries/QueryKeys.ts
@@ -31,6 +31,9 @@ export enum QueryKeys {
   GROUP_INVITE = "groupInvite",
   GROUP_JOIN_REQUEST = "groupJoinRequest",
   PENDING_JOIN_REQUESTS = "pendingJoinRequests",
+
+  // Profiles
+  PROFILE_SOCIALS = "profileSocials",
 }
 
 export const groupsQueryKey = (account: string) => [QueryKeys.GROUPS, account];
@@ -127,3 +130,8 @@ export const pendingJoinRequestsQueryKey = (account: string) => [
   QueryKeys.PENDING_JOIN_REQUESTS,
   account,
 ];
+
+export const profileSocialsQueryKey = (
+  account: string,
+  peerAddress: string
+) => [QueryKeys.PROFILE_SOCIALS, account, peerAddress];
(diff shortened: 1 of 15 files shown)


queries/useProfileSocialsQuery.ts
```diff
@@ -0,0 +1,84 @@
...
+  fetcher: async (addresses: string[]) => {
+    const data = await getProfilesForAddresses(addresses);
+    return data;
+  },
+  resolver: indexedResolver(),
+  scheduler: windowedFiniteBatchScheduler({
+    windowMs: 10,
+    maxBatchSize: 150,
+  }),
+});
```
thierryskoda, line 23: I think react-query handle request deduplication

queries/QueryKeys.ts
```diff
@@ -127,3 +130,8 @@ export const pendingJoinRequestsQueryKey = (account: string) => [
   QueryKeys.PENDING_JOIN_REQUESTS,
   account,
 ];
+
+export const profileSocialsQueryKey = (
+  account: string,
+  peerAddress: string
+) => [QueryKeys.PROFILE_SOCIALS, account, peerAddress];
```
thierryskoda, line 137: Curious, why did you choose to add all keys into 1 file? Just really curious. I'm not doing that but maybe it's the best practice. What I do is put each keys into it's own file

thierryskoda (commented):
Love that! Big fan of react-query. Would just make sure that it's aligned with how @technoplato planned to integrate them so that when we go with XState we don't have to refactor a ton.

queries/useProfileSocialsQuery.ts
thierryskoda: OH very nice yeah makes sense didn't know about this

queries/QueryKeys.ts
alexrisch: I just kinda did it one day, but I think how we are structuring folders by feature it probably makes more sense to split them into the separate files

PR #645, opened 2024-08-31
feat: share frame from context menu

saulmc:

### DIFF
--- a/components/MessageContextMenuWrappers/MessageContextMenuWrapper.tsx
+++ b/components/MessageContextMenuWrappers/MessageContextMenuWrapper.tsx
@@ -76,8 +76,8 @@ export const MessageContextMenuWrapper: FC<MessageContextMenuWrapperProps> = ({
     closeMenu();
   }, [message.content, message.contentFallback, closeMenu]);
 
-  const isFrame = isFrameMessage(message);
   const frameURL = useMemo(() => {
+    const isFrame = isFrameMessage(message);
     if (isFrame) {
       const frames = useFramesStore
         .getState()
@@ -85,7 +85,7 @@ export const MessageContextMenuWrapper: FC<MessageContextMenuWrapperProps> = ({
       return frames[0]?.url;
     }
     return null;
-  }, [isFrame, message.converseMetadata?.frames]);
+  }, [message]);
 
   const contextMenuItems = useMemo(() => {
     const items = [];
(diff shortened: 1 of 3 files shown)


components/MessageContextMenuWrappers/MessageContextMenuWrapper.tsx
```diff
@@ -73,6 +76,17 @@ export const MessageContextMenuWrapper: FC<MessageContextMenuWrapperProps> = ({
     closeMenu();
   }, [message.content, message.contentFallback, closeMenu]);
 
+  const isFrame = isFrameMessage(message);
```
nmalzieu, line 79: this will be recomputed on every render
I think you can move the isFrameMessage method inside the frameURL useMemo and just test if frameURL is not null line 104 since frameURL will always be null if isFrame is false

components/MessageContextMenuWrappers/MessageContextMenuWrapperiOS.ios.tsx
```diff
@@ -35,6 +39,7 @@ const MessageContextMenuWrapperIOSInner: FC<
   );
   const isAttachment = isAttachmentMessage(message.contentType);
   const isTransaction = isTransactionMessage(message.contentType);
+  const isFrame = isFrameMessage(message);
```
nmalzieu, line 42: same

nmalzieu (changes requested):

components/MessageContextMenuWrappers/MessageContextMenuWrapper.tsx
```diff
@@ -82,14 +96,31 @@ export const MessageContextMenuWrapper: FC<MessageContextMenuWrapperProps> = ({
...
       items.push({
-        title: "Copy message",
+        title: "Copy",
         systemIcon: "doc.on.doc",
         onPress: handleCopyMessage,
       });
     }
+    if (frameURL) {
+      items.push({
+        title: "Share",
```
alexrisch, line 106: Don't need to change, for this PR, but can we start putting new features into the translations: i18n/translations/en.ts
and using the translate function

alexrisch (approved):

components/MessageContextMenuWrappers/MessageContextMenuWrapper.tsx
saulmc: For sure. I discovered that function the other night and went to town... https://github.com/ephemeraHQ/converse-app/pull/660

PR #647, opened 2024-09-01
feat: use ephemeral account for deniability

saulmc:
- Updates copy of ephemeral account banner to reflect deniability value prop: _"Disconnect to permanently remove your device from these conversations and ensure deniability."_
- Tap banner to logout and delete
- Only allow disconnect without delete if account is not ephemeral

### DIFF
--- a/hooks/useDisconnectActionSheet.ts
+++ b/hooks/useDisconnectActionSheet.ts
@@ -55,9 +55,3 @@ export const useDisconnectActionSheet = () => {
     [logout, ephemeralAccount]
   );
 };
-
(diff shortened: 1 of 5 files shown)


hooks/useDisconnectActionSheet.ts
```diff
@@ -0,0 +1,63 @@
...
+            method();
+          }
+        }
+      );
+    },
+    [logout, ephemeralAccount]
+  );
+};
+
+export const showDeleteAccountActionSheet = async (
```
alexrisch, line 59: I think this was auto generated

nmalzieu (approved):

Issue #686, opened 2024-09-06
App slows down and explodes when many people are typing


saulmc:
Earlier today the Ephemera channel popped off, and people who were sending messages saw their apps slow down and eventually have to kill them. People who were passively observing didn't really encounter the issue.

https://xmtp-labs.slack.com/archives/C07DX271K6V/p1725580811207959

Alex's assessment: https://xmtp-labs.slack.com/archives/C076F0L877T/p1725641221643879

>Frames
related crashes: [https://converse-app.sentry.io/issues/5806351362/?project=4504757120729088&query=is%3Aun[…]um%5D&referrer=issue-stream&statsPeriod=1h&stream_index=2](https://converse-app.sentry.io/issues/5806351362/?project=4504757120729088&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=2)
Frames re-rendering (not visually)
Error handling
Syncing bad groups
Sending message retries
Typing is slow
Check that we are not doing 2 way conversation across the bridge here
Is there any additional memory leak here?
Many messages
Need to make sure we aren't O(n^2) with some of the data we are requiring per message
Receiving many messages there seems to be issues

PR #693, opened 2024-09-09
A bunch of perf improvements while sending messages

nmalzieu:
Improve perf when sending message, like dependency to context
1st step towards #686


alexrisch (approved):

nmalzieu:
> Looks good, I think some of the useMemos may not be as helpful since it will probably more costly to compute if it should render here's a blog post about it: https://kentcdodds.com/blog/usememo-and-usecallback
> 
> I think we will really need to start looking into the React Compiler as it will reduce a lot of these pains (and probably introduce new ones 🤷)

Yes! The main gain is the context dependency because conversation.lastUpdateAt is updated frequently especially during message sending
The end goal is to make the message cell
- render as little as possible
- be as light as possible => right now we don't really care and put a lot of code/features in it, it's probably not ideal

PR #699, opened 2024-09-10
Update the sign methods in iOS and android

nplasterer:
Updates the methods to hash and uses the identity key directly to sign.

### DIFF
--- a/android/app/src/main/java/com/converse/dev/BackendApi.kt
+++ b/android/app/src/main/java/com/converse/dev/BackendApi.kt
@@ -15,7 +15,6 @@ import android.util.Base64
 import android.util.Base64.NO_WRAP
 
 
-
 private suspend fun makePutRequest(appContext: Context, apiURI: String, address: String, signature: String, jsonObject: JSONObject): ByteArray {
 
     return suspendCancellableCoroutine { continuation ->
--- a/android/app/src/main/java/com/converse/dev/BackendApi.kt
+++ b/android/app/src/main/java/com/converse/dev/BackendApi.kt
@@ -10,7 +10,6 @@ import com.android.volley.toolbox.Volley
 import org.json.JSONObject
 import org.xmtp.android.library.Client
 import org.xmtp.android.library.messages.PrivateKeyBuilder
-import org.xmtp.android.library.Util
 import kotlin.coroutines.resume
 import android.util.Base64
 import android.util.Base64.NO_WRAP
@@ -59,7 +58,7 @@ suspend fun putGroupInviteRequest(appContext: Context, apiURI: String, xmtpClien
     val groupInviteEndpoint = "$apiURI/api/groupJoinRequest/$joinRequestId"
 
     val messageToSign = "XMTP_IDENTITY".toByteArray(Charsets.UTF_8)
-    val signature = Base64.encodeToString(xmtpClient.keys.identityKey.sign(Util.keccak256(messageToSign)).toByteArray(), NO_WRAP)
+    val signature = Base64.encodeToString(xmtpClient.keys.identityKey.sign(messageToSign).toByteArray(), NO_WRAP)
     val address = xmtpClient.address
     val body = JSONObject().apply {
         put("status", status)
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/com/converse/dev/BackendApi.kt
```diff
@@ -55,13 +57,13 @@ private suspend fun makePutRequest(appContext: Context, apiURI: String, address:
 
 suspend fun putGroupInviteRequest(appContext: Context, apiURI: String, xmtpClient: Client, joinRequestId: String, status: String) {
     val groupInviteEndpoint = "$apiURI/api/groupJoinRequest/$joinRequestId"
-    val privateKey = PrivateKeyBuilder.buildFromSignedPrivateKey(xmtpClient.keys.identityKey)
-    val signature = Base64.encodeToString(PrivateKeyBuilder(privateKey).sign("XMTP_IDENTITY".toByteArray()).toByteArray(), NO_WRAP)
+
+    val messageToSign = "XMTP_IDENTITY".toByteArray(Charsets.UTF_8)
+    val signature = Base64.encodeToString(xmtpClient.keys.identityKey.sign(Util.keccak256(messageToSign)).toByteArray(), NO_WRAP)
```
neekolas, line 62: It looks like[ in JS we create a signature proto](https://github.com/ephemeraHQ/converse-app/blob/ededda0108bcd63d7803f942c0cac7df757b666e/utils/xmtpRN/api.ts#L34) and then base64 encode that, not just base64 encoding the raw signature bytes

ios/ConverseNotificationExtension/Xmtp/Client.swift
```diff
@@ -107,11 +107,10 @@ func putGroupInviteRequest(apiURI: String?, account: String, xmtpClient: Client,
     do {
       
       let joinRequestUri = "\(apiURI)/api/groupJoinRequest/\(joinRequestId)"
-      let privateKey = try PrivateKey(xmtpClient.keys.identityKey)
       guard let digest = "XMTP_IDENTITY".data(using: .utf8) else {
         return
       }
-      let signature = try await privateKey.sign(digest).serializedData().base64EncodedString()
+      let signature = try await xmtpClient.keys.identityKey.sign(Util.keccak256(digest)).serializedData().base64EncodedString()
```
neekolas, line 113: Looking at the [JS implementation](https://github.com/ephemeraHQ/converse-app/blob/ededda0108bcd63d7803f942c0cac7df757b666e/utils/xmtpRN/api.ts#L24) (which is what this needs to match if we want it to work), it doesn't seem like we hash the text before signing.

android/app/src/main/java/com/converse/dev/BackendApi.kt
nplasterer: I think this is base61 encoding the signature proto since thats whats returned by the sign function.

neekolas: Oh, weird. OK. Then that's fine

neekolas (approved):

ios/ConverseNotificationExtension/Xmtp/Client.swift
```diff
@@ -107,11 +107,10 @@ func putGroupInviteRequest(apiURI: String?, account: String, xmtpClient: Client,
     do {
       
       let joinRequestUri = "\(apiURI)/api/groupJoinRequest/\(joinRequestId)"
-      let privateKey = try PrivateKey(xmtpClient.keys.identityKey)
       guard let digest = "XMTP_IDENTITY".data(using: .utf8) else {
         return
       }
-      let signature = try await privateKey.sign(digest).serializedData().base64EncodedString()
+      let signature = try await xmtpClient.keys.identityKey.sign(digest).serializedData().base64EncodedString()
```
nplasterer, line 113: Apparently iOS is still not verifying in Javascript. The Android and iOS code is identical and I can validate in a test that the bytes being passed are the same. So there has to be something weird going on here. I'll keep looking into this when I get time.

nplasterer:
I still think this would be good to update from a clean code perspective and example to others. But it doesn't fix the underlying issue so up to you @nmalzieu if you want to merge this or not.

PR #700, opened 2024-09-10
feat: JS Context Menu

alexrisch:
Moved to JS context Menu
Added blur view
Removed old implementation


package.json
```diff
@@ -146,6 +148,7 @@
     "react-native-gesture-handler": "~2.16.1",
     "react-native-get-random-values": "~1.11.0",
     "react-native-gradle-plugin": "^0.71.19",
+    "react-native-hold-menu": "^0.1.6",
     "react-native-ios-context-menu": "^2.5.1",
```
nmalzieu, line 152: this should probably be removed

nmalzieu (approved):
Works pretty well for me on iOS. I'll approve & merge so we can move forward on perf & crashes but there are some small fixes I guess that will need to be done


Small display issue when adding a reaction:
Capture d’écran 2024-09-10 à 12 20 49

PR #706, opened 2024-09-10
fix: update sentry to fix build in xcode beta

saulmc:


ios/Converse.xcworkspace/xcshareddata/swiftpm/Package.resolved
nmalzieu, line 1: this file gets created & removed all the time. Let's not commit it but let's add it to .gitignore

nmalzieu (changes requested):

nmalzieu (approved):

PR #719, opened 2024-09-12
JS context menu polish

saulmc:
- fixes horizontal positioning
- centers reactions display
- increases menu width
- normalizes spacing

### DIFF
--- a/components/Chat/Message/MessageReactionsList.tsx
+++ b/components/Chat/Message/MessageReactionsList.tsx
@@ -310,7 +310,6 @@ const useStyles = () => {
       top: safeAreaInsets.top + 10,
       left: 0,
       right: 0,
-      bottom: 0,
       justifyContent: "center",
       alignItems: "center",
       pointerEvents: "box-none",
--- a/components/Chat/Message/MessageReactionsList.tsx
+++ b/components/Chat/Message/MessageReactionsList.tsx
@@ -11,7 +11,7 @@ import {
   textSecondaryColor,
   tertiaryBackgroundColor,
 } from "@styles/colors";
-import { AvatarSizes, BorderRadius, Paddings } from "@styles/sizes";
+import { AvatarSizes, BorderRadius, Paddings, Margins } from "@styles/sizes";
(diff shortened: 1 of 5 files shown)


components/Chat/Message/MessageReactionsList.tsx
```diff
@@ -210,14 +212,20 @@ const MessageReactionsListInner: FC<MessageReactionsListProps> = ({
...
       {list.length !== 0 ? (
-        <FlatList
-          contentContainerStyle={styles.reactionsContainer}
-          data={list}
-          horizontal
-          renderItem={renderItem}
-          keyExtractor={keyExtractor}
-          showsHorizontalScrollIndicator={false}
-        />
+        <Portal>
```
nmalzieu, line 215: I don't understand the use of Portal here. Isn't Portal used to render somewhere else but it needs a PortalHost?

nmalzieu (changes requested):

components/Chat/Message/MessageReactionsList.tsx
nmalzieu: @saulmc 
Allright not sure I know how this works
Documentation at https://callstack.github.io/react-native-paper/docs/components/Portal/ says
> It requires a [Portal.Host](https://callstack.github.io/react-native-paper/docs/components/Portal/PortalHost) component to be rendered somewhere in the parent tree
and by checking out your branch I don't see PortalHost anywhere
Am I misunderstanding the doc?

saulmc: I think it is included in the PortalProvider in App.tsx. From the docs:

> Note that if you're using the Provider component, this already includes a Portal.Host.

saulmc: Also, I don't really know how this works either -- just copying a pattern that Alex had used for the message bubble.

nmalzieu: got it. Let's go then!

nmalzieu (approved):

PR #730, opened 2024-09-13
Conversation list null states

saulmc:
- Show matchmaker recommendations for new users with onchain recommendations.
- Show QR code for users with no onchain recommendations.
- Show beta chat invite.


lourou:
Looks good @saulmc! 

Two comments:

1. The Share Link is cut off on the iPhone mini due to the reduced screen height on the mini models.

2. When I accept the invite on the web page to join the private beta group, the spinner animates for a few seconds and then displays "An error occurred."

### Screenshots

saulmc:
Thx @lourou , fixed!

config.ts
```diff
@@ -68,7 +68,8 @@ const ENV = {
       (domain) => [`https://${domain}`, `http://${domain}`, domain]
     ),
     enableTransactionFrames: true,
-    betaGroupChatUrl: "https://converse.xyz/group-invite/NfrIwpnhZdRCx2ZyGCg-S",
+    alphaGroupChatUrl:
+      "https://converse.xyz/group-invite/NfrIwpnhZdRCx2ZyGCg-S",
```
cameronvoell, line 72: this url shows up as not found, not sure if it was supposed to be updated as well

cameronvoell (approved):
See comment that theres is one link that didnt seem to be working. But otherwise updates look good. cool improvement!

darickdang:
Here's an initial pass.

I think the "Help shape the future of Converse!" gets lost when stuck at the bottom so I pushed it to the top and made it so it could be dismissed. It may be a useful component to use in the future for other callouts/announcements as well.

#730 - Conversation list null states

PR #732, opened 2024-09-13
Encrypt the converse database, including migrating an existing one

nmalzieu:
Encrypting the converse database 

- generating a salt on device and saving it to keystore
- using that salt + db encryption key when we generate a database
- setup plaintext header of 32 bytes to avoid 0xdead10cc crashes on iOS
- detect an unencrypted database, and migration to encrypt it
- path op-sqlite to be able to handle encrypted database AND unencrypted databases to be able to do the migration


nplasterer (approved):
🔥

PR #735, opened 2024-09-13
Bump to the latest RN version

nplasterer:
Includes the fix for plaintext headers.


ios/Podfile
```diff
@@ -18,7 +18,7 @@ install! 'cocoapods',
 
 # Version must match version from XMTP Podspec (matching @xmtp/react-native-sdk from package.json)
 # https://github.com/xmtp/xmtp-react-native/blob/v2.5.3/ios/XMTPReactNative.podspec#L29
```
cameronvoell, line 20:
```suggestion
# https://github.com/xmtp/xmtp-react-native/blob/v2.5.4/ios/XMTPReactNative.podspec#L29
```

seems the convention is to update this comment as well to match the RN tag

cameronvoell (approved):
looks good!

PR #746, opened 2024-09-17
Style null states according to designs

saulmc:
Reconciles with designs and introduces reusable banner component for future CTAs


lourou (approved):
@saulmc Is there another group invite link? The URL https://converse.xyz/group-invite/NfrIwpnhZdRCx2ZyGCg-S shows a "Group invite not found".

Other than this, everything looks good!

PR #749, opened 2024-09-17
Tap to reveal timestamps

### LINKED ISSUE #696: Add drag to reveal timestamps
Dragging to the left anywhere inside a conversation should reveal message timestamps to the right.

When dragging to the left, messages from me should translate left accordingly, but messages from others should stay in place. This interaction can be seen in iMessage.

lourou:
- Reveal message time on tap, also works when a date is already shown following a date change in message history
- Time displays are localized based on the 12/24-hour formats depending on the phone's locale
- Using Reanimated for animating height, keeping Animate from react native for the existing `<Swipable>` reply animation 

Closes #696

### DIFF
--- a/components/Chat/Message/Message.tsx
+++ b/components/Chat/Message/Message.tsx
@@ -33,7 +33,7 @@ import {
 } from "../../../data/store/accountsStore";
 import { XmtpMessage } from "../../../data/store/chatStore";
 import { isAttachmentMessage } from "../../../utils/attachment/helpers";
-import { getRelativeDate } from "../../../utils/date";
+import { getLocalizedTime, getRelativeDate } from "../../../utils/date";
 import { isDesktop } from "../../../utils/device";
 import { converseEventEmitter } from "../../../utils/events";
 import {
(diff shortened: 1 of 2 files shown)


nplasterer (approved):

components/Chat/Message/Message.tsx
```diff
@@ -133,9 +145,55 @@ const MessageSenderAvatar = ({ message }: { message: MessageToDisplay }) => {
...
-function ChatMessage({ message, colorScheme, isGroup, isFrame }: Props) {
+const ChatMessage = ({
+  account,
+  message,
+  colorScheme,
+  isGroup,
+  isFrame,
+  showTime,
+  toggleTime,
+}: Props & { showTime: boolean; toggleTime: () => void }) => {
```
nmalzieu, line 156: we don't need to pass showTIme & toggleTime, let's define them in that component directly and remove them from the cached component below?

components/Chat/Message/Message.tsx
```diff
@@ -133,9 +145,55 @@ const MessageSenderAvatar = ({ message }: { message: MessageToDisplay }) => {
...
+  const animatedStyle = useAnimatedStyle(() => {
+    return {
+      minHeight: height.value, // Use minHeight for more stable height animations
+      transform: [{ translateY: translateY.value }],
+      opacity: opacity.value,
+    };
+  });
+
+  // Handle animation based on showTime prop
+  React.useEffect(() => {
```
nmalzieu, line 182: this useEffect will be triggered when the component renders then when the props change

when the component renders showTime will be false so the "hide time" animation will trigger but it's not necessary since it's already hidden

not sure if it's really a perf issue but we can  prevent that: add a useRef for the last showtime value starting with `false` and only trigger the show animation if showtime goes from false to true or from true to false

nmalzieu (changes requested):

nmalzieu (approved):

lourou:
Thanks @nmalzieu 🎉

PR #750, opened 2024-09-17
Render emojipicker only once

nmalzieu:
We were rendering the emoji picker once per message. Rendering it only once for the whole convo for increased perf.


lourou (approved):
🔥

PR #764, opened 2024-09-17
Group profile consent, uniform UI Action Sheet

### LINKED ISSUE #592: Group profile consent action still uses "block" and does not alert
In this video, I show how "Block" alerts from the 1:1 profile page but not from the group profile page.

https://github.com/user-attachments/assets/cc6f53ae-4b3a-4f2e-a6e9-ce6db553dba9

Here's a markup with strings we can use, consistent with the "trash can" from the conversation list.

lourou:
Closes #592


components/Chat/ChatPlaceholder/ChatPlaceholder.tsx
```diff
@@ -81,8 +82,7 @@ export default function ChatPlaceholder({ messagesCount }: Props) {
                     options: ["Unblock", "Cancel"],
                     cancelButtonIndex: 1,
                     destructiveButtonIndex: isBlockedPeer ? undefined : 0,
-                    title:
-                      "If you unblock this contact, they will be able to send you messages again.",
+                    title: translate("if_you_unblock_contact"),
```
alexrisch, line 85: 😃

alexrisch (approved):

components/Chat/ChatPlaceholder/ChatPlaceholder.tsx
lourou: @alexrisch thanks for the review 🙏 we've got to translate those strings 😁

PR #777, opened 2024-09-19
feat: default usernames for ephemeral accounts

saulmc:
Thinking up a pseudonymous username is hard. This provides a generated default.

Future: make usernames optional for ephemeral accounts.

### DIFF
--- a/utils/str.ts
+++ b/utils/str.ts
@@ -77,6 +77,14 @@ export const conversationName = (
 export const formatGroupName = (topic: string, groupName?: string) =>
   groupName || capitalize(humanize(topic.slice(14, 46), 3, " "));
 
+export const formatEphemeralUsername = (address: string, username?: string) =>
+  username || humanize(address.slice(2, 42), 2, "");
+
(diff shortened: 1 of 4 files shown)


utils/str.ts
```diff
@@ -77,6 +77,14 @@ export const conversationName = (
 export const formatGroupName = (topic: string, groupName?: string) =>
   groupName || capitalize(humanize(topic.slice(14, 46), 3, " "));
 
+export const formatEphemeralUsername = (address: string, username?: string) =>
+  username || humanize(address.slice(2, 42), 2, "");
+
+export const formatEphemeralDisplayName = (
+  address: string,
+  displayName?: string
+) => displayName || capitalize(humanize(address.slice(2, 42), 2, "", true));
```
alexrisch, line 86: Why is capitalize called here and then also in humanize?

saulmc: Good catch, does not need to be called since the first word will already be capitalized.

lourou (approved):

PR #788, opened 2024-09-20
persist data for inbox filtering right away

### LINKED ISSUE #785: Group requests move between inbox views
Expected:

All group requests are considered spam by default.
Groups will go to the main inbox if:
- You have already granted consent to the group creator
- You have already granted consent to the inviter

Actual:

All group requests start as spam or in the main inbox according to the logic above.
After a period of time, if the user has not accepted the group request, it moves to requests (whether it was originally in spam or the main inbox)

Steps to reproduce:

1. "Hidden requests" move to "you might know"
- Create two ephemeral accounts Alix and Bo
- As Alix, create a group consisting of Alix and Bo
- Switch to Bo. The group invitation will correctly appear under "hidden requests"
- Kill the app, and return to the requests view
- You should now see the group in "you might know"

2. Main inbox group convos move to "you might know"
- As Alix, send Bo a DM to consent to Bo
- Switch to Bo. Create a new group consisting of Bo and Alix.
- Switch back to Alix. The group will correctly appear in the main inbox, because Alix has consented to its creator.
- Kill the app, and return to the requests view
- You should now see the group in "you might know"

saulmc:
Closes #785 - Group requests move between inbox views

- Persist groupCreator and groupAddedBy so that group requests promoted to main inbox stay there
- Compute spam scores for groups so that group requests stay in the proper folder
- Exclude group conversations from likelyNotSpam by default
- Always show the segmented controller, so that spam is not mistaken as requests


data/helpers/conversations/upsertConversations.ts
```diff
@@ -77,10 +81,20 @@ const setupAndSaveConversations = async (
...
   const conversationsToUpsert: Conversation[] = [];
-  conversations.forEach((conversation) => {
+  conversations.forEach(async (conversation) => {
     const alreadyConversationInDbWithTopic =
       alreadyConversationsByTopic[conversation.topic];
 
+    // If spam score is not computed, compute it
+    if (
+      conversation.spamScore === undefined ||
+      conversation.spamScore === null
```
nmalzieu, line 91: are we positive every conversation will get a spam score? We used to not score empty convos but I see that you removed it. Just want to make sure that this won't execute in a loop for some convos. What score will an empty conversation get?

data/helpers/conversations/upsertConversations.ts
```diff
@@ -77,10 +81,20 @@ const setupAndSaveConversations = async (
   });
 
   const conversationsToUpsert: Conversation[] = [];
-  conversations.forEach((conversation) => {
+  conversations.forEach(async (conversation) => {
```
nmalzieu, line 84: using `async` in a `forEach` will not make the `forEach` wait for the `await` 
also this is ou conversation `upsert` method that's supposed to save stuff in the db, I think it should be as fast as possible

data/helpers/conversations/upsertConversations.ts
```diff
@@ -77,10 +81,20 @@ const setupAndSaveConversations = async (
...
       alreadyConversationsByTopic[conversation.topic];
 
+    // If spam score is not computed, compute it
+    if (
+      conversation.spamScore === undefined ||
+      conversation.spamScore === null
+    ) {
+      await computeConversationsSpamScores(account, [
+        conversation as XmtpConversationWithUpdate,
+      ]);
```
nmalzieu, line 95: `computeConversationsSpamScores` can trigger a query to our db (for sender spam score)
not sure we should `await` here because if the network call fails we will not save the conversation to db!

utils/xmtpRN/conversations.ts
```diff
@@ -787,7 +787,9 @@ export const sortRequestsBySpamScore = (
   requests.forEach((conversation) => {
     const isLikelyNotSpam =
       conversation.spamScore !== undefined &&
-      (conversation.spamScore === null || conversation.spamScore < 1);
+      (conversation.spamScore === null || conversation.spamScore < 1) &&
+      conversation.version !== ConversationVersion.GROUP;
+    // @todo => remove this once we have a spam score for groups
```
nmalzieu, line 792: you have effectively implemented a spam score for groups here right? just based on the groupAddedBy spam score?

nmalzieu (changes requested):
A few comments about network / async calls and potential loop, apart from that looks good!

utils/xmtpRN/conversations.ts
saulmc: true! will clarify the comment: we need to think through whether the same scoring logic for DMs should also apply to groupAddedBy senders

data/helpers/conversations/upsertConversations.ts
saulmc: @nmalzieu These diffs on spamScore.ts should make sure every conversation gets scored, right?

https://github.com/ephemeraHQ/converse-app/pull/788/files/39d6fc1b2b170953eb0242b427d57fd86865e3d4#diff-22463bf42246ca396975d8b9e611855052c26a367b7cfc39609b9ee304cfb889L51

https://github.com/ephemeraHQ/converse-app/pull/788/files/39d6fc1b2b170953eb0242b427d57fd86865e3d4#diff-22463bf42246ca396975d8b9e611855052c26a367b7cfc39609b9ee304cfb889R82

cameronvoell: Looks like every conv should get a spam score. Added a debug log here so if we do get in a loop, it will be easy to diagnose. Initial tests looks good!

cameronvoell (approved):
this issue is now looking good for me, nice fix!

https://github.com/ephemeraHQ/converse-app/issues/785

PR #791, opened 2024-09-20
Noe/fix profiles 0x

nmalzieu:
Profiles could show as 0x for 1:1 conversations

Our profiles data was a bit messed up
- we were using mmkv + sqlite persistence but not always updating sqlite. I removed the profile table from sqlite to rely only on mmkv (extensions use mmkv only)
- group members addresses are lowercased in the SDK but 1:1 convos peerAddress is formatted. This can lead to inconsistencies; opened a failing test here to showcase the issue: https://github.com/xmtp/xmtp-react-native/pull/497
- changing Converse code to always check the current address + a lowercased & a formatted version to match all cases
- we were still computing a "conversationTitle" for 1:1 based on current profile data and saving it for reuse in notifications => removing that as now the notification extension can directly access the profile data from mmkv and even refresh it when necessary


nplasterer (approved):

cameronvoell (approved):
code updates look good. I also tested on dev network with an account that had existing v2 conversations and groups. Names showed up as expected.

I created an issue in the RN SDK to normalize address capitalization between V2 and groups to help other who would encounter similar issues => https://github.com/xmtp/xmtp-react-native/issues/498

Nice fix!

PR #795, opened 2024-09-23
update camera and library purpose strings, remove unused permissions

saulmc:
From App Store Review:
"One or more purpose strings in the app do not sufficiently explain the use of protected resources. Purpose strings must clearly and completely describe the app's use of data and, in most cases, provide an example of how the data will be used."

This PR:
- Updates purpose strings to provide the same level of detail as other messaging apps
- Removes permissions that are not currently used by the app

### DIFF
--- a/ios/Converse/Info.plist
+++ b/ios/Converse/Info.plist
@@ -69,15 +69,9 @@
 		<true/>
 	</dict>
 	<key>NSCameraUsageDescription</key>
-	<string>Give $(PRODUCT_NAME) permission to access your camera</string>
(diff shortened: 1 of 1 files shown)


thierryskoda:
I'm probably missing some context here but why aren't we using expo config plugins and put the string in the app.config.ts instead? https://docs.expo.dev/versions/latest/sdk/imagepicker/#example-appjson-with-config-plugin

saulmc:
@thierryskoda does this look alright to you for unblocking the next app store submission?

thierryskoda:
Yes!

PR #796, opened 2024-09-23
add recs back to add group member screen

saulmc:
Unhides matchmaker recommendations from the add group member modal. Many internal testers missed this when it was removed.

### DIFF
--- a/components/Recommendations/Recommendations.tsx
+++ b/components/Recommendations/Recommendations.tsx
@@ -298,6 +298,7 @@ const useStyles = () => {
         default: {
           borderBottomWidth: 0.5,
           borderBottomColor: itemSeparatorColor(colorScheme),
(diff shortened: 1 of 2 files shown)


components/Recommendations/Recommendations.tsx
```diff
@@ -298,6 +298,7 @@ const useStyles = () => {
         default: {
           borderBottomWidth: 0.5,
           borderBottomColor: itemSeparatorColor(colorScheme),
+          paddingLeft: 16,
```
thierryskoda, line 301: a BIG refactor but would be amazing to start using the theme 😄 
`export const Paddings = {
  default: 16,
  small: 8,
  large: 24,
};`

nmalzieu (approved):

PR #797, opened 2024-09-23
fix: Made emoji search a bit better

alexrisch:
Added 100 to hundred keywords
Updated to look in name and also match against emoji
Fixes https://github.com/ephemeraHQ/converse-app/issues/739


thierryskoda:
Maybe we should use something like https://github.com/kentcdodds/match-sorter since we'll probably need to do a lot of filtering like this in the app

lourou (approved):
💯

PR #801, opened 2024-09-23
only show null state if done syncing

### LINKED ISSUE #800: Null state flashes on app open
Null state should only display if conversation loading is complete and there are no conversations.

saulmc:
Closes #800


nmalzieu:
@saulmc Didn't see you had opened this PR, I have also fixed the issue in my sync perf PR.
I'm not sure the `shouldShowConnectingOrSyncing` is what we should test against. This should just be shown if we have no convo at all?  See https://github.com/ephemeraHQ/converse-app/pull/802

PR #802, opened 2024-09-24
Noe/initial sync new perf issues

### LINKED ISSUE #799: unresponsive during initial sync
Slow initial sync blocks UI updates other than the requests count.

Described by Shane on Android below:
> - Was stuck on the null state screen for 15 seconds I thought I lost all my convos
> - I’ve been syncing my inbox for about 5 minutes and it just keeps counting requests
> - I was stuck at 359 requests for 30 seconds and then a few messages finally loaded but all old ones and now I’m just still waiting

Described by another Android tester:
> my inbox is still loading after 20 minutes - the only indication of progress is a slow upcount of my requests.
the app is really slow right now since its loading the inbox.

On iOS, I've encountered a UI lockup while initial loading is taking place, but only for 2-3 seconds.

### LINKED ISSUE #800: Null state flashes on app open

nmalzieu:
With recent changes on spam computation and null states, 2 issues were introduced :

- when we save conversations during the initial sync, we were computing each spam score one by one which was killing perf. Checking agains `initialLoadDoneOnce` prevents that and fixes #799 . The spam score is already computed at the end here: https://github.com/ephemeraHQ/converse-app/blob/main/utils/xmtpRN/sync.ts#L192
- during the initial fetch of conversations list, we were showing the QR code and the user didn't know that we were actually pulling stuff from network. Checking against `initialLoadDoneOnce` will also fix that
- when we open the app and the inbox / requests division has not been computed yet, the qr code would flash. Checking agains total conversations length rather than inbox & requests fixes #800 

N.B. : it means that the null state with QR code will NOT be shown, even during initial sync, for users who have existing conversations. If they have not many conversations, loading them will be fast, so not an issue. If they have many conversations & messages, loading them will take a long time, but I guess they're power users and showing them the null state with QR code is less important


lourou (approved):
Looks great!

PR #815, opened 2024-09-26
perf improvement initial sync android + spam score fixes

nmalzieu:
Fixes for spam score and some android specific requests screen

- some clear spams were shown as non spam (1:1 conversations score was computed before we had the 1st message so we were not considering content)
- killing & reopening the app was making some hidden spam to show in likely not spam => undefined & null were not treated the same
- requests screen for android was freezing if many convos => a ConversationFlashList as a footer of another ConversationFlashList was breaking FlashList optimization. Fixed by showing the toggle as part of the flashlist and removing the Footer

### DIFF
--- a/data/helpers/conversations/spamScore.ts
+++ b/data/helpers/conversations/spamScore.ts
@@ -67,6 +67,7 @@ export const computeConversationsSpamScores = async (
   const sendersSpamScores = await getSendersSpamScores(
     Array.from(conversationsRequesterAddresses)
   );
+
   const topicSpamScores: TopicSpamScores = {};
 
   conversations.forEach((conversation) => {
@@ -80,7 +81,10 @@ export const computeConversationsSpamScores = async (
     const senderSpamScore = sendersSpamScores[senderKey];
     if (
       !conversation.messagesIds.length &&
-      typeof senderSpamScore === "number"
+      typeof senderSpamScore === "number" &&
+      // 1:1 Conversations without messages and no specific sender spam score
+      // should not get a spam score now (waiting for first message)
+      (conversation.isGroup || senderSpamScore !== 0)
     ) {
       // Cannot score an empty conversation further, score is just the
       // sender spam score
--- a/screens/Navigation/ConversationRequestsListNav.tsx
+++ b/screens/Navigation/ConversationRequestsListNav.tsx
@@ -5,6 +5,8 @@ import {
   backgroundColor,
   textPrimaryColor,
 } from "@styles/colors";
+import { ConversationFlatListItem } from "@utils/conversation";
+import { converseEventEmitter } from "@utils/events";
 import React, {
   useCallback,
   useEffect,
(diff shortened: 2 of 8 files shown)


data/helpers/conversations/spamScore.ts
```diff
@@ -80,7 +81,10 @@ export const computeConversationsSpamScores = async (
     const senderSpamScore = sendersSpamScores[senderKey];
     if (
       !conversation.messagesIds.length &&
-      typeof senderSpamScore === "number"
+      typeof senderSpamScore === "number" &&
+      // 1:1 Conversations without messages and no specific sender spam score
+      // should not get a spam score now (waiting for first message)
+      (conversation.isGroup || senderSpamScore !== 0)
```
technoplato, line 87: Hey @nmalzieu - I'm new here so likely lacking context, but the 1:1 in the comments doesn't seem to line up with the check of `.isGroup`. I would imagine that 1:1 conversations would **not** be a group, correct?

Could you help me understand what I'm missing there?

screens/Navigation/ConversationRequestsListNav.tsx
```diff
@@ -151,35 +159,32 @@ export default function ConversationRequestsListNav() {
     <NativeStack.Screen name="ChatsRequests" options={navigationOptions}>
       {(navigationProps) => {
         navRef.current = navigationProps.navigation;
+        let items: ConversationFlatListItem[] = likelyNotSpam;
+        if (likelySpam.length > 0) {
+          items = isSpamToggleEnabled
+            ? [
+                ...likelyNotSpam,
+                {
+                  topic: "hiddenRequestsButton",
```
technoplato, line 168: Is this how we render non "list item" components within the list itself?

nmalzieu: It's a list item, just not a conversation ! But yeah it's a way to render a different type of item in the conversation flash list. We use it only for that specific android toggle but could be useful for other stuff in the future

data/helpers/conversations/spamScore.ts
saulmc: This will leave the conversation spam score null or undefined, which will be treated as spam:

https://github.com/ephemeraHQ/converse-app/pull/815/files#diff-7eda797ad01ac0fefcc17a9a00c4f7fb9607e883f5a65321163a75e96ec1c889R808

nmalzieu: Yes exactly, a 1:1 conversation will be in hidden until we have a first message in it. Most clients create convo then send message almost immediatly, but if we compute spam score too early then we never recompute it.
Also I guess empty 1:1 convos we don't really care about so it's ok if they're in hidden requests?

saulmc: Great. We sometimes see artificially empty 1:1 conversations in the main inbox during initial load/sync. Holding them in hidden requests until they have a message and have been scored makes sense for now.

nmalzieu: Convos I have approved (protocol consent) will still appear in inbox even if no message (or during initial sync).

lourou (approved):

PR #816, opened 2024-09-26
fix: Update statuses for UI

alexrisch:
Moved states into same state object
Added better UI to inform the User of current status of the invite Stop polling on success

### DIFF
--- a/i18n/translations/en.ts
+++ b/i18n/translations/en.ts
@@ -260,6 +260,7 @@ const en = {
   group_join_error: "An error occurred",
   group_join_invite_invalid: "This invite is no longer valid",
   group_finished_polling: "This is taking longer than expected",
+  group_already_joined: "This invite has already been accepted",
 
   // Group Overview
   change_profile_picture: "Change profile picture",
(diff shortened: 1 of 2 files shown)


i18n/translations/en.ts
```diff
@@ -260,6 +260,7 @@ const en = {
   group_join_error: "An error occurred",
   group_join_invite_invalid: "This invite is no longer valid",
   group_finished_polling: "This is taking longer than expected",
```
technoplato, line 262: Since this correlates with `finishedPollingUnsuccessfully`, we should rename the string key to contain that `_unsuccessfully` condition.

technoplato: or `without_being_able_to_join` or something to indicate that this string comes along with some undesirable condition

thierryskoda:
Maybe we could put this into a reusable hook in the same file

alexrisch:
> Maybe we could put this into a reusable hook in the same file

Michael will be refactoring this to a state machine to set a pattern for how we want to set them up in the future

PR #818, opened 2024-09-26
feat: Better Handling with Group invites

alexrisch:
Added handling for Group Ids on Group requests
Navigate the user to the screen when already joined
Inform the user the invite is no longer valid if they have been removed/group is not active

https://github.com/user-attachments/assets/b1aa14cd-e669-4b9d-a6c3-64ba134bc476


technoplato:
Awesome, thanks for sharing the video.

technoplato (approved):

PR #819, opened 2024-09-26
Conversation context menu

### LINKED ISSUE #727: Conversation context menu
Replicate the context menu long press interaction on conversations as implemented in iMessage, Messenger, WhatsApp, and Telegram.

- [x] Long press a conversation in the conversation list to reveal a context menu (replaces long-press to pin)
- [x] Add "Pin" to context menu
- [x] Add "Mark as Unread" to context menu
- [x] Add "Delete" to context menu
- [x] Display auxiliary preview of the conversation

lourou:
Conversation context menu implementation with auxiliary preview

Closes #727

### DIFF
--- a/components/ConversationContextMenu.tsx
+++ b/components/ConversationContextMenu.tsx
@@ -6,6 +6,9 @@ import {
   AUXILIARY_VIEW_MIN_HEIGHT,
   HOLD_ITEM_TRANSFORM_DURATION,
   contextMenuStyleGuide,
+  BACKDROP_DARK_BACKGROUND_COLOR,
+  BACKDROP_LIGHT_BACKGROUND_COLOR,
+  ITEM_WIDTH,
 } from "@utils/contextMenu/constants";
 import { BlurView } from "expo-blur";
 import React, { FC, memo, useCallback, useEffect } from "react";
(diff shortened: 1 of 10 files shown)


components/ConversationContextMenu.tsx
```diff
@@ -0,0 +1,230 @@
...
+  useSharedValue,
+  withTiming,
+} from "react-native-reanimated";
+
+import { useIsSplitScreen } from "../screens/Navigation/navHelpers";
+
+const AnimatedBlurView =
+  Platform.OS === "ios"
+    ? Animated.createAnimatedComponent(BlurView)
+    : Animated.createAnimatedComponent(View);
```
thierryskoda, line 41: Should we make a BlurView component in `components` that we can reuse. So if later BlurView is fixed on Android we can just change it in one file

thierryskoda (approved):
Small comment but the rest looks good to me!

thierryskoda:
Missing some context maybe but also what we could do is in Chat create a components just for the messages so we can display that instead of the whole Chat component. Might help with performance.

components/ConversationContextMenu.tsx
lourou: Yes, good one, just pushed this change

lourou:
> Missing some context maybe but also what we could do is in Chat create a components just for the messages so we can display that instead of the whole Chat component. Might help with performance.

I've started working on a Chat component refactor on a separate branch/PR, so we can improve perf on the auxiliary preview.

PR #823, opened 2024-09-26
wip: group join refactor - xstate / dependency control

technoplato:
I promise I will clean up this PR
clean-it-up-mcjuggernuggets

this pr is going to contain a whole bunch of different ideas

it will be broken down into smaller prs after initially sharing with team and determining what we like/don't like. 

GitHub conversation: https://github.com/ephemeraHQ/converse-app/discussions/929


technoplato:
* **#823** <a href="https://app.graphite.dev/github/pr/ephemeraHQ/converse-app/823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @technoplato and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

i18n/translations/en.ts
```diff
@@ -260,13 +260,23 @@ const en = {
   // Group Invites
   group_admin_approval:
     "A group admin may need to approve your membership prior to joining.",
+  "A group admin may need to approve your membership prior to joining.":
```
alexrisch, line 263: This feels like a weird approach, why use full sentences rather than keys?

openai/openrouter/openai/o1-preview
alexrisch, line 1: This can be deleted

queries/useGroupInviteQuery.ts
```diff
@@ -8,7 +8,11 @@ export const useGroupInviteQuery = (groupInviteId: string) => {
   const account = useCurrentAccount() as string;
 
   return useQuery<GroupInvite | undefined>({
+    // I need to learn more about how query keys work to determine
```
alexrisch, line 11: Query keys are just a way to select a certain query status
Technically it can be anything, but it's normally best to combine the dependencies of the function

First we want to have a dependency on the group invite id, so if you look at a different group invite id, it will have a different state

Since the QueryClient is app wide, this means we should also set query keys by account since if you look at the group invite id as one account and then switch to another, it should be a completely different query

queries/useGroupInviteQuery.ts
```diff
@@ -8,7 +8,11 @@ export const useGroupInviteQuery = (groupInviteId: string) => {
...
+    // the significance of this query key with an account
     queryKey: groupInviteQueryKey(account, groupInviteId),
+    // does this need to be in useQuery for any
```
alexrisch, line 14: Nope doesn't need to be

Think of useQuery and really all of the query client usage more as a network state wrapper.
The queryFn is just the actual function within the wrapper

I think the majority of XState usage would not actually interact with much of the state of react query other than maybe some prefetches

XState should be our necessary consistent state
React Query should be the helpful things to load later on

If we zoom out a little from this instance and instead look at initial app load
Xstate should handle:
1. Loading the client/checking if that is correct/usable
2. Loading our necessary information (Conversations, Groups, Consent)
3. Optionally it can tell our queries to prefetch some data (like fetching profiles of the first 20 items in our conversation list)

React Query:
React Query handles the actual decision of fetching from the network when needed

If XState was also in charge of fetching all profiles it can become pretty messy and slow

But if React Query handles it, from a DX experience it's a lot cleaner to say:
useProfileData and rely on the caching/persistence offered by the library

Now if we want to look at this use case, we _can_ use React Query just for initial loading of what the group invite network data is, but it's really not providing much other than syntactic sugar on top of useState/useEffect

So I think what makes sense is to remove useQuery entirely from this case and manage the one time load in XState

features/GroupInvites/GroupLink.tsx
```diff
@@ -1,23 +1,22 @@
+import Button from "@components/Button/Button";
```
alexrisch, line 1: We can just remove this whole screen I believe

features/GroupInvites/GroupInvite.screen.tsx
```diff
@@ -1,81 +1,88 @@
...
   );
+
   const { allowGroup } = useGroupConsent(newGroup?.topic || "");
   const handlingNewGroup = useRef(false);
 
   const handleNewGroup = useCallback(
     async (group: GroupWithCodecsType) => {
-      if (group.client.address === account && !handlingNewGroup.current) {
+      const whatDoesThisConditionIndicate =
+        group.client.address === account && !handlingNewGroup.current;
```
alexrisch, line 74: This is checking if the new group is the correct account of what's expected from the user
There's also a check to make sure it's not doing multiple groups
I think both are a bit hacky and can probably be removed

technoplato: Thanks for answering all my questions.

Two follow ups here: 

1) what does `group.client` mean? 

2) When could `client.address` NOT equal the current account?

i18n/translations/en.ts
technoplato: @thierryskoda and I discussed this approach and while at first glance it may seem strange, you should look at it from the perspective of the screen you're building / looking for a screen from a design/app run.

If I try and create a short key for the string, I have to search and find it in this file, and then jump to where that string is used in the code. That's more steps and those steps add up quick since string search is easiest way to find where in the app something is happening. 

cc @thierryskoda for thoughts

openai/openrouter/openai/o1-preview
technoplato: Thanks, not sure how that got here at all. No sweat on reviewing this anymore since it's still in draft by the way! Don't want to waste your time w stuff like this.

queries/useGroupInviteQuery.ts
technoplato: That totally makes sense - thanks for the explanation. Went through `react-query` docs and they were super helpful, as well as the source. It's pretty clean!

queries/useGroupInviteQuery.ts
technoplato: @alexrisch Let me get a little bit more of the code done and we can review it and continue iterating for what should live where from there. It's a little too tricky without my end ready for review first. 

TLDR: 

`xstate` will DEFINITELY not reinvent the wheel that `react-query` has built, so querying/caching/optimistic updating logic will live in `react-query` and `xstate` will react to, observe, and trigger things in `react-query` land. 

Initially, I am going to keep `zustand` as a layer to share state as the transition to `xstate` completely as a state _management_ layer will be more lengthy and there will always be times where we may not want it just to share state in place A and place B. This is a conversation we will continue to revisit to make our workflows as ergonomic as possible. 

> So I think what makes sense is to remove useQuery entirely from this case and manage the one time load in XState

Agree here, although I'm also a huge fan of consistency. SHOULD our entire client -> network communication be done through `queryClient`? My initial gut says yes and that's how I'll build this, but we can always revisit. 

Having consistency to me is more valuable than saving a few keystrokes, and it will be a valuable low risk exercise to understand how to interact with network via `queryClient` in `xstate` within the Controlled Dependencies layer. 

Lots of words, code will help! Stay tuned 📺

i18n/translations/en.ts
thierryskoda: Agree with Michael that this way worked best for me. Especially for finding screen.

features/GroupInvites/GroupInvite.screen.tsx
alexrisch: There is a client for each connected account, but I don't think that's really a case in this code, so I think it should be removed

features/Inspection/useEventTargetShim.ts
technoplato, line 1: Required for remote inspector for state machines (stately sky) in expo. Not used yet but keeping it here so it's ready.

i18n/i18n.ts
```diff
@@ -17,6 +17,48 @@ i18n.translations = {
...
+ * // Using English sentences as keys:
+ * i18n.t("Hello, how are you?") // Works as-is
+ * i18n.t("menu::Edit::Cut") // Nested structure
+ *
+ * @sideEffects
+ * - Changes how nested keys are accessed in translations
+ * - Existing translation key strings may need updating
+ * -- Regex to find foo.bar.strings: translate\(.*\..*"\)
+ */
+// export const TranslationNestedSeparator = "::" as const;
```
technoplato, line 57: saving this for later

jest.config.ts
technoplato, line 1: moved config from `package.json#jest` into `jest.config.ts` and added settings to ensure `env` variables were available at test time

jest.setup.ts
```diff
@@ -46,10 +47,10 @@ jest.mock("react-native-fs", () => {
     touch: jest.fn(),
     MainBundlePath: jest.fn(),
     CachesDirectoryPath: jest.fn(),
-    DocumentDirectoryPath: jest.fn(),
```
technoplato, line 49: `string#join` was complaining that we were passing it a function, not a string

package.json
```diff
@@ -61,6 +61,7 @@
     "@sentry/react-native": "^5.31.1",
     "@shopify/flash-list": "1.6.4",
     "@stardazed/streams-polyfill": "^2.4.0",
+    "@statelyai/inspect": "^0.4.0",
```
technoplato, line 64: not used yet but will be

queries/groupConsentMutationUtils.ts
```diff
@@ -0,0 +1,109 @@
...
+export type GroupConsentAction = "allow" | "deny";
+
+export type GroupConsentMutationProps = {
+  account: string;
+  topic: string;
+  groupId: string;
+  action: GroupConsentAction;
+};
+
+export const createGroupConsentMutationObserver = (
```
technoplato, line 22: first stab at using `react-query` imperatively

queries/useGroupsQuery.ts
```diff
@@ -1,18 +1,20 @@
...
 
 import { groupsQueryKey } from "./QueryKeys";
 import { entify, EntityObject } from "./entify";
 import { queryClient } from "./queryClient";
-import {
-  ConverseXmtpClientType,
-  GroupWithCodecsType,
-} from "../utils/xmtpRN/client";
 
-type GroupMembersSelectData = EntityObject<GroupWithCodecsType, string>;
```
technoplato, line 13: string is default 2nd generic, can be removed

queries/useGroupsQuery.ts
```diff
@@ -1,18 +1,20 @@
...
+type GroupMembersSelectData = EntityObject<GroupWithCodecsType>;
 
-const groupsQueryFn = async (account: string) => {
+export const groupsQueryFn = async (account: string): Promise<GroupsEntity> => {
```
technoplato, line 17: prefer explicit type

screens/Navigation/SplitScreenNavigation/SplitRightStackNavigation.tsx
```diff
@@ -109,9 +103,8 @@ export default function SplitRightStackNavigation({
           {WebviewPreviewNav()}
           {ProfileNav()}
           {GroupNav()}
-          {GroupLinkNav()}
```
technoplato, line 112: Group link nav was not needed, `GroupInvite` replaced by `JoinGroup` to be more explicit

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "expo/tsconfig.base",
   "compilerOptions": {
```
technoplato, line 3: required for dynamic imports in `clients`. Will be used to avoid needlessly mocking unused transitive dependencies

utils/api.test.ts
```diff
@@ -28,7 +28,8 @@ jest.mock("./keychain", () => {
   };
 });
 
-describe.skip("GroupInviteLink", () => {
+// todo(lustig): incorporate these tests into JoinGroupClient#live
+describe("GroupInviteLink", () => {
```
technoplato, line 32: fixed by new `jest.config.ts` see comment there for more details

utils/api.ts
```diff
@@ -17,7 +18,7 @@ import type { ProfileSocials } from "../data/store/profilesStore";
 import type { Frens } from "../data/store/recommendationsStore";
 import { getXmtpApiHeaders } from "../utils/xmtpRN/api";
 
-const api = axios.create({
+export const api = axios.create({
```
technoplato, line 21: used to create `JoinGroupClient.live`

technoplato:
closing draft as this was implemented via stacks!

Issue #836, opened 2024-09-27
Emoji picker should only require one tap


saulmc:
When I have to search for an emoji to use in a reaction, and the emoji picker goes into full mode, I will have to tap on the emoji twice.

The first time will minimize the emoji picker into the collapsed mode, the second time will add the emoji as the reaction 

[Slack Message](https://xmtp-labs.slack.com/archives/C07DX271K6V/p1727446118439609)

PR #837, opened 2024-09-28
fix: Added handled for persisting taps

alexrisch:
Dismiss keyboard when emoji is actually tapped
Closes https://github.com/ephemeraHQ/converse-app/issues/836

### DIFF
--- a/components/EmojiPicker/EmojiRowList.tsx
+++ b/components/EmojiPicker/EmojiRowList.tsx
@@ -27,6 +27,9 @@ interface EmojiRowListProps {
 
 const keyExtractor = (_: unknown, index: number) => String(index);
 
+// Works around issue with Android not picking up scrolls
+const ListRenderer = Platform.OS === "ios" ? ReanimatedFlashList : FlatList;
+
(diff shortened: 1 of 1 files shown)


components/EmojiPicker/EmojiRowList.tsx
```diff
@@ -27,6 +27,9 @@ interface EmojiRowListProps {
 
 const keyExtractor = (_: unknown, index: number) => String(index);
 
+// Works around issue with Android not picking up scrolls
+const ListRenderer = Platform.OS === "ios" ? ReanimatedFlashList : FlatList;
```
thierryskoda, line 31: Yeah I also had many issues with FlashList on Android

thierryskoda (approved):

components/EmojiPicker/EmojiRowList.tsx
alexrisch: There's definitely a trade off with it unfortunately, we might want to keep a tracking list to make better decisions about if we feel it's worthwhile keeping it

Issue #386, opened 2024-07-26
Bug: Frames not posting to the button `post_url`


neekolas:
### Describe the bug

In the [Open Frames standard](https://github.com/open-frames/standard?tab=readme-ov-file#determining-the-post_url), there is a hierarchy of which URL a button click should post to. If the button itself has a `target` or `post_url` defined that should be used instead of the frame-level post_url. It seems that our implementation isn't doing that.

### Expected behavior

Button clicks should always URL as defined in the standard

PR #838, opened 2024-09-28
fix: Frames Post Handling

alexrisch:
Updated to include button post url to follow frames spec

Closes https://github.com/ephemeraHQ/converse-app/issues/386

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,9 +95,8 @@ yarn dev
 yarn test
 ```
 
+# Frames
 
-# Key Cases Needed for App Release
+Frames are expected to follow the Open Frames Standard https://github.com/open-frames/standard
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -95,9 +95,8 @@ yarn dev
 yarn test
 ```
 
+# Frames
+
+Frames are expected to follow the Open Frames Standard https://github.com/open-frames/standard
 
-# Key Cases Needed for App Release
```
alexrisch, line 99: Removed this since it's not entirely accurate right now any ways, I think it's good to be open about release blockers when dealing with open source, but since it's not currently accurate it's not worth having it here right now

nmalzieu (approved):

PR #849, opened 2024-09-30
Chat component refactor

lourou:
Refactor Chat component and optimize message handling for chat preview (long press on chat > auxiliary preview)

- Remove the `readOnly` props in the chat component, and create a dedicated, more lightweight `<ChatPreview />` component instead
- Extract `usePeerSocials`,  `renderItem` and `getItemType`
- Add optional `lastMessages` parameter in `getListArray` to limit the number of returned messages in `<ChatPreview />`
- Remove `readOnly` prop from `Chat` component, simplifying logic for padding and focus states
- Replace `ConverseChat` with the newly refactored `Chat` component
- Remove default export
- Remove unused code in `screens/ConversationReadOnly.tsx` (onLeaveScreen)

Linked to: https://github.com/ephemeraHQ/converse-app/pull/819#issuecomment-2379231954

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -241,8 +241,31 @@ const getListArray = (
   return reverseArray;
 };
 
+const useAnimatedListView = (
+  conversation: XmtpConversationWithUpdate | undefined
+) => {
+  // The first message was really buggy on iOS & Android and this is due to FlashList
+  // so we keep FlatList for new convos and switch to FlashList for bigger convos
+  // that need great perf.
+  return useMemo(() => {
+    const isConversationNotPending = conversation && !conversation.pending;
+    return isConversationNotPending && Platform.OS !== "web"
+      ? ReanimatedFlashList
+      : ReanimatedFlatList;
+  }, [conversation]);
+};
+
+const useIsShowingPlaceholder = (
+  listArray: MessageToDisplay[],
+  isBlockedPeer: boolean,
+  conversation: XmtpConversationWithUpdate | undefined
+): boolean => {
+  return listArray.length === 0 || isBlockedPeer || !conversation;
+};
+
 export function Chat() {
   const conversation = useConversationContext("conversation");
+  const AnimatedListView = useAnimatedListView(conversation);
   const isBlockedPeer = useConversationContext("isBlockedPeer");
   const onReadyToFocus = useConversationContext("onReadyToFocus");
   const transactionMode = useConversationContext("transactionMode");
@@ -342,8 +365,11 @@ export function Chat() {
   ]);
 
   const framesStore = useFramesStore().frames;
-  const showPlaceholder =
-    listArray.length === 0 || isBlockedPeer || !conversation;
+  const showPlaceholder = useIsShowingPlaceholder(
+    listArray,
+    isBlockedPeer,
+    conversation
+  );
   const renderItem = useRenderItem(
     xmtpAddress,
     conversation,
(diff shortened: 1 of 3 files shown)


components/Chat/Chat.tsx
```diff
@@ -343,36 +391,7 @@ export default function Chat({ readOnly = false }: { readOnly?: boolean }) {
...
-          item.converseMetadata.attachment.size.height
-        ).toFixed(2);
-        return `ATTACHMENT-${aspectRatio}-${fromMeString}`;
-      } else {
-        return `${item.contentType}-${fromMeString}`;
-      }
-    },
-    [framesStore]
-  );
+  const itemType = getItemType(framesStore);
```
nmalzieu, line 394: weird naming as `itemType` is still a method here

components/Chat/Chat.tsx
```diff
@@ -460,6 +478,112 @@ export default function Chat({ readOnly = false }: { readOnly?: boolean }) {
   );
 }
 
+// Lightweight chat preview component used for longpress on chat
+export function ChatPreview() {
```
nmalzieu, line 482: I like that we cleaned the component and extracted some hooks & methods from it but still there is a lot of copy/paste here. why do we feel a new component is better than a readOnly parameter? do we need to maintain both?

nmalzieu (changes requested):
Nice! Just a small question about duplicated code

components/Chat/Chat.tsx
thierryskoda: Agree that doing optional props or using composable pattern can have a fine line.

I also don't often like composable when we know that most of the things we extract we know we won't reuse them many time in the app.

But I think for the animation performance issue it's still worth it to do it here.

It's also easier to understand and isolate component in that way instead of having so many different ways a component can render and react.

Optional props are really not recommended for that. Especially in big component.

I also think that 1-2 lines hooks like those for example are perfectly fine and actually perfect way to make our component composable.

```
const peerSocials = usePeerSocials();
const isSplitScreen = useIsSplitScreen();
```
 
We could even add more to make sure that if we change something in our chat we only change it once by adding:

```
function useAnimatedListView(conversation) {
  const conversationNotPendingRef = useRef(
    conversation && !conversation.pending
  );

  const AnimatedListView =
    conversationNotPendingRef.current && Platform.OS !== "web"
      ? ReanimatedFlashList
      : ReanimatedFlatList;

  return AnimatedListView;
}
```

And 

```
function useIsShowingPlaceholder(listArray, isBlockedPeer, conversation) {
  return listArray.length === 0 || isBlockedPeer || !conversation;
}
```

Just my opinion. Happy to chat more about that.

alexrisch: Further I think that needs to be memoized in some way, I'm pretty sure 
```
const AnimatedListView =
    conversationNotPendingRef.current && Platform.OS !== "web"
      ? ReanimatedFlashList
      : ReanimatedFlatList;
```
Is causing a large bit of rerendering since it's a new component each time

lourou: Thanks @nmalzieu @thierryskoda and @alexrisch – I've just pushed an update with more custom hooks, also memoized the `AnimatedListView`

components/Chat/Chat.tsx
```diff
@@ -241,8 +241,31 @@ const getListArray = (
...
+  }, [conversation]);
+};
+
+const useIsShowingPlaceholder = (
+  listArray: MessageToDisplay[],
+  isBlockedPeer: boolean,
+  conversation: XmtpConversationWithUpdate | undefined
+): boolean => {
+  return listArray.length === 0 || isBlockedPeer || !conversation;
+};
```
thierryskoda, line 264: I like that when there's more than 1 argument we use object arguments instead. But my own opinion. Not sure if we have some specifics docs around that.

thierryskoda: For a hook like that I would do this. (again just my own opinion lol)

```
const useIsShowingPlaceholder = (args: {
  listArray: MessageToDisplay[];
  isBlockedPeer: boolean;
  hasConversation: boolean;
}): boolean => {
  return args.listArray.length === 0 || args.isBlockedPeer || ! args.hasConversation;
};
```

lourou: Hey @thierryskoda

Thanks for your feedback. As you suggested, I've refactored `useIsShowingPlaceholder` and `useRenderItem` to use the object parameter pattern. I did direct destructuring in the function parameters rather than using 'args'.

Cheers!

nmalzieu (approved):

PR #855, opened 2024-09-30
fix: Remove Alpha Banner

alexrisch:
Remove Alpha banner


components/Chat/ChatNullState.tsx
```diff
@@ -68,7 +60,7 @@ const ChatNullState: React.FC<ChatNullStateProps> = ({
 
   return (
     <View style={styles.container}>
-      {!hasUserDismissedBanner && (
+      {/* {!hasUserDismissedBanner && (
```
alexrisch, line 63: Commented out as we may want to include a similar link soon, just not the alpha group chat one

nmalzieu (approved):

PR #868, opened 2024-10-01
fix: infinite app reloading in development using expo

thierryskoda:
* This was causing infinite reloading in development when using `expo start`
* Cleaned it
* Removed the desktop related stuff

Still need some refactor but at least fix the infinite reloading for now


thierryskoda:
Did some tests and everything worked 🙏

screens/Navigation/Navigation.tsx
```diff
@@ -98,10 +99,15 @@ export const navigationAnimation = Platform.OS === "ios" ? "default" : "none";
 export default function Navigation() {
   const colorScheme = useColorScheme();
   const splashScreenHidden = useAppStore((s) => s.splashScreenHidden);
+
   return (
     <GestureHandlerRootView style={{ flex: 1 }}>
       <NavigationContainer
-        linking={splashScreenHidden ? (linking as any) : undefined}
+        linking={splashScreenHidden ? linking : undefined}
+        onReady={() => {
```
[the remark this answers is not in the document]
thierryskoda: Yes! I'll do it in my navigation refactor if it's ok. Since I'm moving lots of things related to it

screens/Navigation/Navigation.tsx
```diff
@@ -98,10 +99,15 @@ export const navigationAnimation = Platform.OS === "ios" ? "default" : "none";
...
+          logger.debug("Navigation ready");
+          useAppStore.setState({ navigationReady: true });
```
nmalzieu, line 109: please use `setNavigationReady`

components/StateHandlers/InitialStateHandler.tsx
```diff
@@ -25,84 +26,106 @@ const getSchemedURLFromUniversalURL = (url: string) => {
...
 
 export default function InitialStateHandler() {
   const colorScheme = useColorScheme();
-  const setDesktopConnectSessionId = useOnboardingStore(
-    (s) => s.setDesktopConnectSessionId
-  );
-  const { setSplashScreenHidden, hydrationDone } = useAppStore(
-    useSelect(["setSplashScreenHidden", "hydrationDone"])
+
+  const url = Linking.useURL();
```
nmalzieu, line 36: using useURL that way I'm not sure you will handle all cases because it will "remember" the last deeplink forever
Example : 
- be logged into the app on (only) one account
- click on a dm deeplink / universal link => opens converse on the page of the convo
- logout from the account without killing the app
- log back it to the account (or another account with profile already setup)
- you will be redirected to the convo while you clearly didn't just click on the link! That's because the system remembers that deeplink

It's not a very big issue for DMs (but still annoying) but could be for other stuff (I remember desktop login had the issue)

nmalzieu (changes requested):
Thanks!

- If I'm not mistaken the `isDevelopmentClientURL` is enough to prevent the infinite app reloading?
- Not sure the desktop stuff should be removed - as soon as the JS SDK has v3 support (few weeks tops) we will want to re-enable desktop login, wdyt? Desktop login stuff needs to handle deep links even when user is not logged in.

components/StateHandlers/InitialStateHandler.tsx
thierryskoda: Normally this should be okay because the url won't change so won't trigger the useEffect below.

BUT you're right that we're also listening to check if the user is signed in so this can make the useEffect retrigger. Will push a fix

nmalzieu: @thierryskoda the flow I described is one I experienced! useEffect below will be triggered the `isSignedIn` will go to true then false then back to true because you logout and relogin.

Issue #857, opened 2024-09-30
Allow me to create a group with just myself


saulmc:
Right now there is no way to make a group by yourself. This makes it difficult to obtain a group invite link to invite people to join your group.

Proposed short-term solution is to display "Next" to the New group screen even if no members are selected:

[Slack Message](https://xmtp-labs.slack.com/archives/C07DX271K6V/p1727711356641019?thread_ts=1727710890.824399&cid=C07DX271K6V)

PR #869, opened 2024-10-01
feat: Enable Solo Group

alexrisch:
Allow group creation when 0 members are selected
Closes https://github.com/ephemeraHQ/converse-app/issues/857


technoplato (approved):
lgtm - any protocol things to be aware of here or was this always an app layer restriction?

PR #877, opened 2024-10-01
Refactor: navigation

thierryskoda:


App.tsx
```diff
@@ -115,13 +117,15 @@ export default function App() {
...
                 }
               >
-                <PortalProvider>
-                  <View style={styles.safe}>
-                    <XmtpEngine />
-                    <Main />
-                    <DebugButton ref={debugRef} />
-                  </View>
-                </PortalProvider>
+                <GestureHandlerRootView style={{ flex: 1 }}>
```
alexrisch, line 120: Want to include this from a stylesheet?

components/StateHandlers/InitialStateHandler.tsx
```diff
@@ -25,6 +25,10 @@ const getSchemedURLFromUniversalURL = (url: string) => {
   return schemedURL;
 };
 
+const isDevelopmentClientURL = (url: string) => {
```
alexrisch, line 28: Want to make a ticket to fix this stuff up a bit more?

I think there's some issues with the current integration

App.tsx
thierryskoda: @alexrisch The reason I didn't do it (and also didn't in a lot of other places) is because I feel we should talk about how we want to handle styles in the app. There are many ways. For example, I've always put the styles inline. I feel it's much faster to develop that way. And usually if we do it right, we have amazing composability so there are not that many inline styles that pollutes everything

components/StateHandlers/InitialStateHandler.tsx
thierryskoda: Oh! Like this is causing other issues? Or the reloading is still happening anyway?

App.tsx
alexrisch: @thierryskoda doesn't this pose rerender issues?
Isn't each style object then seen as a new object?

thierryskoda: Yes it will rerender you're right! Often my strategy tho is not worry about those because this component shouldn't rerender anyway.

But happy to make it clean and put it in memoized style object instead. Just need to figure out the way we'll do this. (useStyle?, createStyleSheet? objects like Ignite is doing? Tailwind class style?).

alexrisch:
Some testing:

1. Navigating to Conversation Screen from New Conversation screen causes a red error screen: Error: NativeViewGestureHandler must be used as a descendant of GestureHandlerRootView. Otherwise the gestures will not be recognized. See https://docs.swmansion.com/react-native-gesture-handler/docs/installation for more details.
2. Router.goBack doesn't seem to do anything on iPad (I'm running iPad for the first time now anyways)
3. Can make several accounts at once during onboarding by navigating back (tested using ephemeral), I think we want to use replace on the screen rather than navigate once the account is created

alexrisch:
4. image
Profile shows as a modal on iPad, previously was in the right navigation
I'm ok with that personally, but other issue is then "modify" shows in the right navigation

alexrisch:
@thierryskoda have a PR to your branch to fix some of these things: https://github.com/ephemeraHQ/converse-app/pull/1009

I think it over all looks good, the only issues I am having all seem related to:
1. Navigation from the Profile screen for Android and the SplitScreen navigator
2. Back button with the new modal back button 

These issues are on Android and Splitscreen, iOS is working perfectly with

thierryskoda:
Closing because now we're in feature/refactor-navigation

PR #878, opened 2024-10-01
feat: Self Group Hack

alexrisch:


utils/xmtpRN/conversations.ts
```diff
@@ -203,7 +203,8 @@ const handleNewConversation = async (
     isGroup &&
     !(conversation as GroupWithCodecsType).members.some(
       (m) => m.addresses[0].toLowerCase() === client.address.toLowerCase()
-    );
+    ) &&
+    (conversation as GroupWithCodecsType).creatorInboxId !== client.inboxId;
```
nmalzieu, line 207: why do we need to add this if we have empty groups exactly? Also is that temporary shouldSkip check still useful?

nmalzieu (approved):
Approved with one question!

utils/xmtpRN/conversations.ts
alexrisch: @nmalzieu Probably not useful, let me test without that fix, could clean up some of this code

technoplato: @nmalzieu can we create empty groups right now? I was under the impression we could not. (asking for a different use case)

alexrisch: @technoplato the only thing stopping us from creating empty groups is the app layer, protocol layer supports it

Issue #809, opened 2024-09-25
Reactions quick menu sometimes crowds the attribution panel


saulmc:
![image](https://github.com/user-attachments/assets/d99ca937-c369-4c21-bbbd-34717e006b22)

PR #890, opened 2024-10-02
Reply to long messages

### LINKED ISSUE #767: Cannot reply to longer messages from context menu
From Peter: When I long press reply on a longer post, I cannot ever get to the keyboard, and thus can never reply.

image

lourou:
Closes #767 and #809


components/Chat/Message/MessageContextMenu.tsx
```diff
@@ -101,14 +101,32 @@ const BackdropComponent: FC<{
     const getTransformValue = () => {
       if (itemRectY.value > AUXILIARY_VIEW_MIN_HEIGHT + safeAreaInsets.top) {
         const spacing = 16;
-        const topTransform =
+        const statusBarHeight =
+          Platform.OS === "android" ? StatusBar.currentHeight ?? 0 : 0;
```
thierryskoda, line 105: could be cool to centralize this logic inside a layout.ts file in theme or styles or constants or idk haha

thierryskoda (commented):
For some of the math calculations, it could be nice to have some comments explaining where the values come from and why. 

Here's an example. Not the best but gives you an idea.

```
export function getReducedProfileMinHeight() {
  return (
    BOTTOM_SHEET_HANDLE_BAR_HEIGHT +
    getFirstRowHeight() +
    USER_PROFILE_SPACE_BETWEEN_ROW +
    getMutualFriendsRowHeight() +
    USER_PROFILE_SPACE_BETWEEN_ROW +
    // CTA
    getButtonHeight("md") +
    (bottomNotchHeight || 0) +
    // Because we don't want the CTA to be too close to the bottom
    theme.space[4]
  )
}
```

if someone came into my code, seeing `theme.space[4]` could be confusing. So I added some comments above. I also extracted all into a function the tells me what is the height we're getting there.

lourou:
@thierryskoda added comments, constants and some fixes

PR #891, opened 2024-10-02
chore: Update Version Number

alexrisch:
Update version number
use shared version number instead of platform


nmalzieu (changes requested):
this app.json file is imported in multiple places in the codebase and `appJson.expo.ios.version` / `appJson.expo.android.version` are referenced

nmalzieu (approved):

Issue #908, opened 2024-10-03
Deeplinks for groups & ability to prefill text


nmalzieu:

alexrisch:
This is included in 2.0.5

PR #921, opened 2024-10-04
Noe/text input deeplink

nmalzieu:
Implementing #908 

Goal is to be able to
- deeplink to a group knowing its group id, & optionally prefill the text input
- deeplink to a 1:1 knowing the peer address (already implemented!) & optionally prefill the text input (new!)
- prefill the text input for the currently opened conversation without knowing anything about it (from a frame for instance)

This PR does the following:

- introduces a more robust deeplink method parsing for conversations, to parse various arguments (`peer`, `groupId`, `text`)
- renames the existing `message` parameter to `text` on Conversation params
- ability to have converse deeplinks (converse://) in frames
- ability to click dm universal links (https://converse.xyz/dm/...) in app without being redirected to safari
- new group deeplink if we know the group id, redirecting to the group and telling the user if the group is unknown (not a part of the group)

Effectively, here are the converse scheme deeplinks that now work (adapt with converse-dev:// to test locally)

- `converse://dm?peer=0xno12.eth` to open / create the main 1:1 convo with 0xno12.eth. Already existed!
- `converse://dm?peer=0xno12.eth&text=hello` to open / create the main 1:1 convo with 0xno12.eth & prefill text with string `hello`
- `converse://group?groupId=f7349f84925aaeee5816d02b7798efbc` to navigate to group with id `f7349f84925aaeee5816d02b7798efbc` locally if it exists
- `converse://group?groupId=f7349f84925aaeee5816d02b7798efbc&text=hello` to navigate to group with id `f7349f84925aaeee5816d02b7798efbc` locally if it exists and prefill text with string `hello`
- `converse://?text=hello` to fill the text input of the currently active convo

Here are the universal links that are also introduced (adapt with https://dev.converse.xyz to test locally)

- `https://converse.xyz/dm/0xno12.eth` to open / create the main 1:1 convo with 0xno12.eth. Already existed!
- `https://converse.xyz/dm/0xno12.eth?text=hello` to open / create the main 1:1 convo with 0xno12.eth & prefill text with string `hello`. Also the website has been adapted in [that PR](https://github.com/ephemeraHQ/converse-website/pull/15/files) to pass the `text` parameter to the converse deeplink if it got opened in Safari for instance

However some universal links do not exist yet

-  `https://converse.xyz/?text=hello` because we cannot have a universal link on the root `/` because it would prevent smartphone users from navigating to our home page
- `https://converse.xyz/group/f7349f84925aaeee5816d02b7798efbc?text=hello` to navigate to a group page (with or without `text`). This one should work as a universal link (to automatically open converse if clicked) but we are unable to create a matching webpage (in case it is opened in Safari) because we don't have information on the group from the group id. We would need to present an anonymous group page with a button with a custom scheme link, but I think we can do that later

So for Frames cc @fabriguespe , the recommendation would be to use custom scheme deeplinks rather than universal links for a while (these actions would be best with custom open frames actions I think anyway, to work cross client)

### DIFF
--- a/components/Chat/Frame/FramePreview.tsx
+++ b/components/Chat/Frame/FramePreview.tsx
@@ -149,16 +149,21 @@ export default function FramePreview({
   const onButtonPress = useCallback(
     async (button: FrameButtonType) => {
       if (button.action === "link") {
+        if (!button.target) return;
         const link = button.target;
-        if (
-          !link ||
-          !(
-            link.startsWith("http") || link.startsWith(`${config.scheme}://`)
-          ) ||
-          !(await Linking.canOpenURL(link))
-        )
+        try {
+          const url = new URL(link);
+          if (
+            (url.protocol === "http:" ||
+              url.protocol === "https:" ||
+              url.protocol === `${config.scheme}:`) &&
+            (await Linking.canOpenURL(link))
+          ) {
+            Linking.openURL(link);
+          }
+        } catch {
           return;
-        Linking.openURL(link);
+        }
         return;
       }
       if (!conversation) return;
--- a/components/Chat/Frame/FramePreview.tsx
+++ b/components/Chat/Frame/FramePreview.tsx
@@ -152,7 +152,9 @@ export default function FramePreview({
         const link = button.target;
         if (
           !link ||
-          !link.startsWith("http") ||
+          !(
+            link.startsWith("http") || link.startsWith(`${config.scheme}://`)
+          ) ||
           !(await Linking.canOpenURL(link))
         )
           return;
--- a/components/StateHandlers/InitialStateHandler.tsx
+++ b/components/StateHandlers/InitialStateHandler.tsx
@@ -4,26 +4,15 @@ import * as Linking from "expo-linking";
 import { useCallback, useEffect, useRef } from "react";
 import { useColorScheme } from "react-native";
 
-import config from "../../config";
 import { useAppStore } from "../../data/store/appStore";
 import { useOnboardingStore } from "../../data/store/onboardingStore";
 import { useSelect } from "../../data/store/storeHelpers";
 import {
+  getSchemedURLFromUniversalURL,
   navigateToTopicWithRetry,
   topicToNavigateTo,
 } from "../../utils/navigation";
 import { hideSplashScreen } from "../../utils/splash/splash";
-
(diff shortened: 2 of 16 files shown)


fabriguespe:
@jhaaaa this would be great to add in the developers section of Converse

components/Chat/Frame/FramePreview.tsx
```diff
@@ -152,7 +152,9 @@ export default function FramePreview({
         const link = button.target;
         if (
           !link ||
-          !link.startsWith("http") ||
+          !(
+            link.startsWith("http") || link.startsWith(`${config.scheme}://`)
```
technoplato, line 156: I've been burned in the past with manually string matching URLs. Could we add the react-native-url-polyfill and perform parsing with it instead?

components/StateHandlers/InitialStateHandler.tsx
```diff
@@ -4,27 +4,16 @@ import * as Linking from "expo-linking";
 import { useCallback, useEffect, useRef } from "react";
 import { useColorScheme } from "react-native";
 
-import config from "../../config";
 import { useAppStore } from "../../data/store/appStore";
 import { useOnboardingStore } from "../../data/store/onboardingStore";
 import { useSelect } from "../../data/store/storeHelpers";
 import {
+  getSchemedURLFromUniversalURL,
```
technoplato, line 11: what is a schemed url?

nmalzieu: A custom scheme like converse:// and not a http link. Would you prefer another naming?

screens/Navigation/navHelpers.ts
```diff
@@ -5,31 +5,115 @@ import {
...
 import config from "../../config";
 import { isDesktop } from "../../utils/device";
 
-export const getConverseStateFromPath = (path: string, options: any) => {
-  // dm method must link to the Conversation Screen as well
-  // but prefilling the parameters
-  let pathForState = path;
-  if (Platform.OS === "web" && pathForState.startsWith("/")) {
-    pathForState = pathForState.slice(1);
+export const getConverseStateFromPath =
```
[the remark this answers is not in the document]
nmalzieu: done!

alexrisch (approved):

PR #924, opened 2024-10-04
feat: Performance Tests

alexrisch:
Added Performance tests
Closes https://github.com/ephemeraHQ/converse-app/issues/923


nmalzieu (approved):
@alexrisch  Nice!

What does `danger` do exactly?
Also, do these perf test fail in some cases or it's just measuring stuff? Are we notified if perf drops, do we have specific threshold per component?

alexrisch:
Will merge once tests pass
*Performance regression tests will not pass this first time as it checks against main, which does not have tests or reassure

alexrisch:
> @alexrisch Nice!
> 
> What does `danger` do exactly? Also, do these perf test fail in some cases or it's just measuring stuff? Are we notified if perf drops, do we have specific threshold per component?

Dangerfile is a way to automate some PR review, it may be overkill in our case, but it's suggested by reassure so stuck with their suggested approach and can refine later
 
The new actions should test any file with ".perf-test"
When that's done it will test against a baseline of the file in the .reassure directory, but that's gitignored right now (again part of the suggested approach that maybe doesn't make sense for our use case)

To run tests we write we can use ```yarn reassure```

PR #927, opened 2024-10-04
feat: Pressable Group Updates 

alexrisch:
Added handling when pressing a display name in the group updated messages


i18n/translations/en.ts
```diff
@@ -311,6 +311,15 @@ const en = {
     "Share your feedback and ideas with the Converse team and other alpha testers.",
   joinAlphaGroup: "Join Converse VIP Alpha Chat",
   linkCopied: "Link copied",
+
+  // Group Updated Message
+  group_name_changed: '{{name}} changed the group name to "{{newValue}}".',
+  group_member_joined: "{{name}} joined the conversation",
```
saulmc, line 317: @covrter This string has always felt that it's missing an explanation of how they got into the group. I think WhatsApp/Telegram/Signal all use variations of "{{inviter}} added {{joiner}}" for direct adds and "{{joiner}} joined via invite link". Should we do this too?

covrter: I think that's valuable to include the inviter, because it communicates trust — do we have the data handy?

It does again raise the question of when the "join" happens — if I'm added by a non-consented contact "badboy", I would not expect to see "badboy added andrew" until I consent to join the convo. I might expect for "badboy invited andrew" to appear in the chat, and then when I consent "andrew joined" to appear. Does that feel right?

Either way this is its own task :)

Telegram example for my future reference
image

components/PressableProfileWithText.tsx
```diff
@@ -0,0 +1,48 @@
+import { navigate } from "@utils/navigation";
+import { memo, useCallback, useMemo } from "react";
+import { StyleProp, TextStyle } from "react-native";
+import ParsedText from "react-native-parsed-text";
+
+const PressableProfileWithTextInner = ({
+  profileAddress,
+  profileDisplay,
+  text,
+  textStyle,
```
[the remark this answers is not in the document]
alexrisch: Updated to include them, but right now the styling would match the nonpressable text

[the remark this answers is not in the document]
saulmc: Thanks! The styling is just font weight bold: https://xmtp-labs.slack.com/archives/C076F0L877T/p1727901469220339

i18n/translations/en.ts
alexrisch: We do have the information readily available for who added who (whom?)
But the other members of the group do not know if/when a different user "approves" (currently labeled as joins) the group

saulmc: just riffing here, and this would be out of scope for now, but could the approval consent message be presented as a group update message?

components/PressableProfileWithText.tsx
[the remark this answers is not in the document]
alexrisch: @saulmc Updated!
Simulator Screenshot - iPhone 15 Plus - 2024-10-04 at 15 54 04

thierryskoda:
I just want to ensure these UI/UX changes are P0 or maybe P1, since we’ll redo most of it later with the design system.

Also, the tests will likely break after refactoring, making things harder to refactor. So my suggestion would be to not do any tests for now.

i18n/translations/en.ts
alexrisch: Do you mean the consent popup? "Do you want to join this group?"

saulmc: Yea, when the user taps the join button, that triggers a consent message right? Maybe it could also trigger a group update message?

alexrisch:
@thierryskoda I'll stop with component tests, but we need to find a balance between new features and introducing the design system

alexrisch:
@thierryskoda Updated with the design system:
Simulator Screenshot - iPhone 15 - 2024-10-15 at 14 20 51
I think we'll need to make a few updates to the presets to handle this though

i18n/translations/en.ts
alexrisch: It triggers a consent message, but that's on a different topic that other users in the group could not decrypt themselves (I cannot see your consent preferences)

The app _can_ trigger some custom content type, but I think that's getting close to concerns with privacy

saulmc: Ah, right. What is the privacy concern with using a content type message?

alexrisch: Exposing consent of a group to the entire rest of the group.
If you "accept" a group, and then the app sends a message to tell everyone you have consented to the group might be a little weird

Unless we avoid showing the Group Updated messages and only show the Converse Group Consent message

saulmc: I'm probably missing something, but that seems desirable. Group membership is already public to group members, and group consent creates a useful division between "invited" and "joined". We should check how other apps handle.

/cc @covrter

alexrisch: yeah we would need to be mindful of how we are labelling these things

If Converse is relying on that, we need to run through all previous messages to see who's "consented" to the group, which will likely not be a complete list unless you are the creator of the group

thierryskoda (commented):
Check this loom before reading my comments https://www.loom.com/share/c4ebb553b56d48ad9fef405144f21fbc

i18n/translations/en.ts
covrter: >  Group membership is already public to group members, and group consent creates a useful division between "invited" and "joined". We should check how other apps handle.

Altho I don't quite follow the technical details … this resonates. I imagine two posts in the channel:
1. Alex invited Saul (when Alex "adds" Saul)
2. Saul joined (Saul consents)

Related riff, and forgive me if we already do this:

When I'm invited but haven't consented, I imagine
- i appear in member list as Invited
- i don't count towards members total #

When I'm invited and decline the invitation
- i don't appear in member list

thierryskoda (approved):

PR #928, opened 2024-10-04
chore: Update Build Scripts

alexrisch:
Added App Release Action
Removed schedule on internal action
Increment app version for next build

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,8 +1,8 @@
 {
   "expo": {
-    "version": "2.0.3",
+    "version": "2.0.4",
     "ios": {
-      "buildNumber": "4"
+      "buildNumber": "1"
     },
     "android": {
       "versionCode": 205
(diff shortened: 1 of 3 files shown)


app.json
```diff
@@ -1,8 +1,8 @@
 {
   "expo": {
-    "version": "2.0.3",
+    "version": "2.0.4",
     "ios": {
-      "buildNumber": "4"
+      "buildNumber": "1"
```
thierryskoda, line 5: I think Apple might complain because they always want a buildNumber higher than what was previously released 😮

thierryskoda (approved):

app.json
thierryskoda: I love using auto increment for that https://docs.expo.dev/build-reference/app-versions/#remote-version-source

thierryskoda: But maybe we can add this later

alexrisch: buildNumber needs to be higher based on the appVersion, so it should always increase the app version first, and then reset this to "1"

Koleok: yeah actually with that option set, you may be able to get rid of the extra line in the version bump script.

PR #949, opened 2024-10-08
feat: Notify When Group Join Fails

alexrisch:
When silent handling of group join requests failed then notify the user

Closes https://github.com/ephemeraHQ/converse-app/issues/982

### DIFF
--- a/screens/GroupInvite.tsx
+++ b/screens/GroupInvite.tsx
@@ -17,10 +17,6 @@ import {
   getGroupJoinRequest,
   GroupJoinRequestStatus,
 } from "@utils/api";
-import {
-  getInviteJoinRequest,
-  saveInviteJoinRequest,
-} from "@utils/groupInvites";
 import { getTopicFromGroupId } from "@utils/groupUtils/groupId";
 import logger from "@utils/logger";
 import { GroupWithCodecsType } from "@utils/xmtpRN/client";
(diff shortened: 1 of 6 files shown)


nmalzieu (approved):
code looks clean
@alexrisch what is it handling exactly?
- it shows a notification to the admin if failure to add to a group? what is the content of the notification? is there a deeplink to the group page?
- do we know when this usually fails? did we ever log it? or you suspect this is issue we have when Joining... is too long?
- should there be an Android part as well?

alexrisch:
@nmalzieu It's a good idea to add the deep link to the group, it helps the user accept easier, so will convert to draft to add that

screens/GroupInvite.tsx
```diff
@@ -100,24 +92,20 @@ export default function GroupInviteScreen({
...
+    const groupsBeforeJoining = await fetchGroupsQuery(account);
+    if (groupId && groupsBeforeJoining.byId[getTopicFromGroupId(groupId)]) {
+      // User has already been added to the group
+      handleNewGroup(groupsBeforeJoining.byId[getTopicFromGroupId(groupId)]);
+      return;
+    }
     logger.debug(
       `[GroupInvite] Before joining, group count = ${groupsBeforeJoining.ids.length}`
     );
-    let joinRequestId = getInviteJoinRequest(account, groupInvite?.id);
```
technoplato, line 109: was using an old `inviteJoinRequest` perhaps the culprit for people's joins timing out do you think?

if not - why was this removed in favor of always creating one?

alexrisch: 2 fold:
1. If a user presses again it will actually send the notification again, so has a chance to potentially reduce the manual step
2. Moved this check to the backend relevant PR is here: https://github.com/ephemeraHQ/converse-backend/pull/39

PR #958, opened 2024-10-09
fix: Syncing

alexrisch:
Added retryWithBackoff function
Added wait util
Added tests for both
Updated exported sync function to use retryWithBackoff Updated stream calls to retry if failed, but not attempt another retry Removed onSyncLost in sync function and move to an outer function to be called


nmalzieu (approved):
LGTM!

@alexrisch does it fix the issue you had in the android simulator? the initial app freeze / unresponsiveness?

PR #963, opened 2024-10-10
feat: add theme + start of design system + haptics

thierryskoda:


design-system/Text/Text.presets.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type IPresets =
+  | "default"
+  | "bold"
+  | "heading"
+  | "subheading"
+  | "formLabel"
+  | "formHelper";
+
+export const $presets: Record<IPresets, ThemedStyleArray<TextStyle>> = {
```
technoplato, line 14: What's the $ prefix for? Not used to seeing that outside the scope of observables / jQuery

utils/haptics.ts
```diff
@@ -0,0 +1,148 @@
...
+const impactFeedbackStyle = ExpoHaptics.ImpactFeedbackStyle;
+
+const notificationFeedbackStyles = ExpoHaptics.NotificationFeedbackType;
+
+const selectionAsync = () => {
+  try {
+    // logger.debug("selectionAsync");
+    ExpoHaptics.impactAsync(impactFeedbackStyle.Light).catch(sentryTrackError); // Selection doens't seem to work? So use light impact for now
+  } catch (error) {
+    sentryTrackError(error);
```
alexrisch, line 15: Goes for all of these, is there a world where this would ever get hit? Or just planning for the future if additional things get added in here?

alexrisch (approved):

utils/haptics.ts
thierryskoda: you're right it's overkill. removed!

design-system/Text/Text.presets.ts
thierryskoda: Removed!

It's how InfiniteRed do styling with Ignite.

So that it's easier for auto complete for example you type $ and see all the styling available.

I removed it for now until we all agree on this way of working.

I don't want this to class with other things in the app

PR #970, opened 2024-10-10
fix: Performance Improvements for Conversation Screen

alexrisch:
General Performance Cleanup for Conversation Screen Moved hardcoded arrays out of react context
Moved functions out of context
Memoize context value
useMemo on useStyles

### DIFF
--- a/components/Chat/Chat.tsx
+++ b/components/Chat/Chat.tsx
@@ -273,6 +273,8 @@ const useIsShowingPlaceholder = ({
   return messages.length === 0 || isBlockedPeer || !conversation;
 };
 
+const keyExtractor = (item: MessageToDisplay) => item.id;
+
 export function Chat() {
   const conversation = useConversationContext("conversation");
   const AnimatedListView = useAnimatedListView(conversation);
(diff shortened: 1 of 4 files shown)


components/Chat/Chat.tsx
```diff
@@ -606,36 +606,40 @@ export function ChatPreview() {
...
-      zIndex: 0,
-    },
-    inChatRecommendations: {
-      borderBottomWidth: 0.5,
-      borderBottomColor: itemSeparatorColor(colorScheme),
-      marginHorizontal: 20,
-      marginBottom: 10,
-    },
-  });
+  return useMemo(
```
nmalzieu, line 609: I wonder if that does anything? The useStyles hook only already depends on the useColorScheme hook so is only executed when this one changes? But I might be wrong and it executes every time the Chat component rerenders?

nmalzieu (approved):
Can be merged but I was wondering if the useMemo was actually doing something? I guess you saw some perf improvement with that?

components/Chat/Chat.tsx
alexrisch: @nmalzieu it was updating on every rerender!
So it's likely best to use useMemo in all useStyles usage

PR #984, opened 2024-10-12
feat: Android Variants

alexrisch:
Added Android Variants
Moved folder structures
Removed Android build scripts used to update new variant info Aligned eas.json profiles
Added new manifest and strings for Android resource merger


android/app/src/main/google-services.json
```diff
@@ -0,0 +1,39 @@
+{
```
thierryskoda, line 1: This file shouldn't be in Git 😵

alexrisch: Yeah, it currently is though, will need to look into how to keep these more secure, but if someone REALLY wanted to I believe they could get this info from the compiled app

nplasterer (approved):

PR #986, opened 2024-10-15
fix: Top level performance improvements

alexrisch:
Performance improvements for App component
Moved constants out of React Context
Moved listeners inside of providers
Small useMemo usage


lourou (approved):
Looks good!

PR #1026, opened 2024-10-17
Remove Android Drawer

alexrisch:
Removes Android Drawer


nmalzieu (approved):
LGTM! Is it temporary or did we decide that we will definitely not have a left drawer on Android?
cc @covrter @saulmc

thierryskoda (approved):

PR #1036, opened 2024-10-19
fix: Xmtp Engine Rerenders, Race Conditions, Crashes 

alexrisch:
Refactored Xmtp Engine to be outside of React Context
Refactored Xmtp Cron to be outside of React Context
Adds subscriptions
Moves app state into folder and adds new app state utils

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -26,7 +26,7 @@ import { Provider as PaperProvider } from "react-native-paper";
 import { ThirdwebProvider } from "thirdweb/react";
 
 import "./utils/splash/splash";
-import XmtpEngine from "./components/XmtpEngine";
+import { XmtpCron } from "./components/XmtpEngine";
 import config from "./config";
 import { useAppStore } from "./data/store/appStore";
 import { useSelect } from "./data/store/storeHelpers";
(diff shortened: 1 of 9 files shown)


technoplato:
Hmm, does this have any concrete indication of performance improvement or does it make the code any more testable?

I just want to make sure if we are going to make structural/architectural refactors, we do so with purpose and planning. (Not saying this doesn't by the way and I was out Friday so missed any potential conversation there)

This just seems like a HUGE core piece of our app and we need to make sure it's architected to be as robust and testable as possible.

nmalzieu:
Agreed!
This looks clean and I understand that we prefer that kind of singletons rather than the component
But since we don't have tests yet it could be a bit dangerous
+ the XmtpEngine was a component so indeed it might rerender but I wouldn't say it was slow to rerender? Or did you experience a slow rerender of that one?

For that kind of "refactoring out of the blue" that is not part of a specific issue, I think it would help the rest of the team to explain
- why we do it
- if we experience improvement by doing it (basically telling us : before I had that, now I don't)?

But I would be down to merge it anyways, and it is definitely more testable.

Koleok (dismissed):
I think there is a time and a place where a singleton makes sense, and this is it :)

PR #1052, opened 2024-10-22
Reactions drawer

### LINKED ISSUE #938: Reactions detail drawer

lourou:
Closes #938 - [Reactions detail drawer](https://github.com/ephemeraHQ/converse-app/issues/938)


.github/workflows/build-app-release.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         run: node scripts/build/android/build.js
 
       - name: Build for Android
-        run: eas build --profile production --platform android --non-interactive --auto-submit --no-wait
+        run: eas build --profile production-android --platform android --non-interactive --auto-submit --no-wait
```
lourou, line 57: @alexrisch I'm not sure about this. I think I introduced it this morning during a rebase/merge conflict.

App.tsx
```diff
@@ -142,9 +144,13 @@
...
-                  <PortalProvider>
-                    <App />
-                  </PortalProvider>
+                  <GestureHandlerRootView>
+                    <BottomSheetModalProvider>
+                      <PortalProvider>
+                        <App />
+                      </PortalProvider>
+                    </BottomSheetModalProvider>
+                  </GestureHandlerRootView>
```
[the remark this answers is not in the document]
lourou: @thierryskoda should we do this? I'm not quite sure :)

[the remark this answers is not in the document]
thierryskoda: @lourou Oh yes we should! :)

theme/palette.ts
```diff
@@ -25,6 +25,9 @@ export const lightPalette = {
   edge: "rgba(255, 255, 255, 0.15)",
 
   ...paletteShared,
+
+  danger: "#FF3B30",
```
alexrisch, line 29: What's the difference between danger and red?
Also since these are the same I think they should go into the shared palette
cc @thierryskoda

scripts/build/build.js
```diff
@@ -77,7 +77,7 @@ const build = async () => {
   const buildCommand = "eas";
   const buildProfile =
     env === "production" || env === "preview"
-      ? `${env}${buildInternalProduction ? "-internal" : ""}`
+      ? `${env}-${platform}${buildInternalProduction ? "-internal" : ""}`
```
alexrisch, line 80: I don't think this should be included

design-system/BottomSheet/BottomSheetModal.tsx
```diff
@@ -40,8 +43,15 @@ export const BottomSheetModal = memo(
...
           }
+          enableDynamicSizing={false} // By default we don't want enable dynamic sizing
           backdropComponent={BottomSheetBackdropOpacity}
           handleComponent={renderHandleComponent}
+          backgroundStyle={[
+            {
+              backgroundColor: theme.colors.background.raised,
+            },
+            backgroundStyle,
+          ]}
```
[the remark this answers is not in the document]
alexrisch: @lourou this is a good suggestion from coderabbit

components/Chat/Message/MessageReactions/MessageReactionsDrawer/MessageReactionsDrawer.tsx
```diff
@@ -0,0 +1,163 @@
...
+                <Avatar
+                  size={theme.avatarSize.md}
+                  uri={item.reactor.avatar}
+                  name={item.reactor.userName}
+                />
+                <Text style={styles.userName}>{item.reactor.userName}</Text>
+                <Text style={styles.reactionContent}>{item.content}</Text>
+              </HStack>
+            ))}
+        </BottomSheetScrollView>
```
thierryskoda, line 126: curious how big the groups we will have but if we end up have LOTS of reactions we might need to use a FlashList here. but we can keep for later maybe.

thierryskoda (approved):
Looks great to me! here the few things I saw that maybe we would want to change but maybe not also if we have other priorities.

* Using useStyles, maybe we would want to use the $ pattern I showed
* Using FlashList instead of .map
* Separating reactions and list items to memoized them for better rerender performance

theme/palette.ts
thierryskoda: Maybe ask Andrew if adding a new "red" color was intended, otherwise yes we can just reuse the original red we had

design-system/BottomSheet/BottomSheetModal.tsx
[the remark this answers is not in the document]
thierryskoda: This component should **never** rerender so I think it's overkill but we can do it do make the code cleaner I don't mind :)

components/Chat/Message/MessageReactions/MessageReactionsDrawer/MessageReactionsDrawer.tsx
lourou: @thierryskoda I've upgraded to `FlashList` to account for the potential many reactions we'll need to display in the drawer :) 🍿  https://cln.sh/K4HgZc6swdw2skX2Z4tK

lourou:
Thanks @thierryskoda for the review! I create separate issues below: 
- https://github.com/ephemeraHQ/converse-app/issues/1144
- https://github.com/ephemeraHQ/converse-app/issues/1143

.github/workflows/build-app-release.yml
lourou: updated to we don't have this diff anymore

App.tsx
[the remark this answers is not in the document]
lourou: done ✅

scripts/build/build.js
lourou: removed ✅

PR #1074, opened 2024-10-22
feat: TextField design system

thierryskoda:
* Showed a couple example

Simulator Screenshot - iPhone 16 Pro - 2024-10-22 at 19 49 45
Screenshot 2024-10-22 at 7 46 46 PM


design-system/Button/OldButton.tsx
technoplato, line 1: can we delete this file?

alexrisch (approved):

PR #1076, opened 2024-10-23
feat: fix icon button styling

thierryskoda:


design-system/IconButton/IconButton.example.tsx
```diff
@@ -0,0 +1,47 @@
+import { memo } from "react";
+
+import { useAppTheme } from "../../theme/useAppTheme";
+import { HStack } from "../HStack";
+import { VStack } from "../VStack";
+import { IconButton } from "./IconButton";
+
+export const IconButtonExample = memo(function IconButtonExample() {
```
alexrisch, line 8: Is Storyboard worthwhile? Or some sort of example screen?

alexrisch (approved):

design-system/IconButton/IconButton.example.tsx
thierryskoda: Every time I added Storyboard in my app it was always overkill. Even when we were 10. So for now I'm just doing those so that if ever we need to see we can just import those.

Might find a better solution later or even delete them TBH.

PR #1093, opened 2024-10-24
fix: internal preview build iOS

thierryskoda:
@alexrisch  this is a small fix to fix what happened here with my internal build on iOS https://github.com/ephemeraHQ/converse-app/actions/runs/11499980582/job/32011918449

I think there's a lot of clean up that could be done for deployment but didn't want to put more time now until you think it could be a huge gain.

I know you also have some work in progress on that.

## Summary by CodeRabbit

- **Chores**
	- Removed the GitHub Actions workflow for creating TestFlight builds for iOS, which eliminates automated build submissions.
	- Simplified the `eas.json` configuration by removing `extends` properties for iOS and Android configurations, while adding specific `ascAppId` values for iOS.


alexrisch:
@thierryskoda going to close this out since it should be resolved with the other changes

PR #1138, opened 2024-10-29
fix: onboarding connect wallet

thierryskoda:
* https://xmtp-labs.slack.com/archives/C07NSHXK693/p1730215871169219?thread_ts=1730197942.805269&cid=C07NSHXK693
* I still think it would help if you go over the list of QA I proposed last time for onboarding
For instance one tricky case I had listed was : connect a walletconnect wallet, do a first signature for XMTP login, go back (top left arrow), try again => it will probably fail in the current branch!


components/Onboarding/ConnectViaWallet/useInitConnectViaWalletState.ts
```diff
@@ -10,12 +10,16 @@ import {
...
 import { useConnectViaWalletContext } from "./ConnectViaWallet.context";
+import { ensureError } from "../../../utils/error";
 import { thirdwebClient } from "../../../utils/thirdweb";
 import { isOnXmtp } from "../../../utils/xmtpRN/client";
 import { getInboxId } from "../../../utils/xmtpRN/signIn";
 
-// For now let's keep Thirdweb and the hooks because I haven't found a better way to do it.
-// But ideally a lot of this it outside React.
+/**
+ * For now let's keep Thirdweb and the hooks because I haven't found a better way to do it.
```
alexrisch, line 19: Yeah there's probably some way to hook into these without react, but not sure how much is being exposed.

It's using react-query under the hood, but the query client is not exposed
Maybe some way to patch package it if we really need to

navigation/useNavigation.tsx
```diff
@@ -1,97 +1,74 @@
...
+  useRoute as useRouteNavigation,
 } from "@react-navigation/native";
-import { useEffect, useMemo } from "react";
+import { NativeStackNavigationProp } from "@react-navigation/native-stack";
+import { useEffect } from "react";
 
 import { NavigationParamList } from "../screens/Navigation/Navigation";
 
+// Extend global namespace for type safety
+declare global {
```
alexrisch, line 13: Should this go in a types.d.ts file instead?

I guess command clicking does bring you to the definition

alexrisch (approved):

PR #1150, opened 2024-10-31
feat: Better control over our logger

thierryskoda:
While debugging connecting wallet I was annoyed with all the logs we had for other modules.

So I created this so we can create specific logger for each big module.

That way, we can enable/disabled different modules logs using env variables


alexrisch:
Do you find this more useful than just filtering in whatever debugger (Flipper etc.)?

PR #1157, opened 2024-11-01
feat: React Native 75

alexrisch:
Updates for React Native 75


alexrisch:
iOS is working
Android was getting an error locally so checking with CI

PR #1164, opened 2024-11-05
feat: design system snackbar + design system examples

thierryskoda:
* snackbar component and service to show them
* remove index.ts in /theme
* fix all places where we were using the index.ts of /theme


package.json
```diff
@@ -135,6 +137,7 @@
     "i18n-js": "3.9.2",
     "ipfs-car": "^1.0.0",
     "libphonenumber-js": "^1.10.49",
+    "lodash": "^4.17.21",
```
alexrisch, line 140: Why was lodash added? I feel like I missed the usage somewhere

thierryskoda: You're right! I'm used to always use it in my project, so maybe I had a thought of adding it but yes ending up not using it. Removed!

components/Snackbar/Snackbar.example.tsx
```diff
@@ -0,0 +1,91 @@
+import { showSnackbar } from "@components/Snackbar/Snackbar.service";
+import { Button } from "@design-system/Button/Button";
+import { VStack } from "@design-system/VStack";
+import { useAppTheme } from "@theme/useAppTheme";
+import { Image } from "expo-image";
+import { memo } from "react";
+import { Dimensions } from "react-native";
+
+export const SnackbarExample = memo(function SnackbarExample() {
```
alexrisch, line 9: ❤️

components/__tests__/__snapshots__/Button.test.tsx.snap
```diff
@@ -66,10 +66,13 @@ exports[`Button Component Default (Android) Button renders correctly with danger
         undefined,
         [
           {
-            "0": [Function],
```
alexrisch, line 69: Whoops 😄

alexrisch (approved):

PR #1167, opened 2024-11-06
feat: Handle V3 DMs when ready 

alexrisch:
Added helpers
Updated naming to match v3 rather than just groups
Updated Pinned Conversations and Conversation List


components/PinnedConversations/PinnedV3Conversation.tsx
```diff
@@ -1,61 +1,25 @@
...
+    return null;
+  }
+  if (conversationIsGroup(conversation)) {
     return (
-      <GroupAvatarDumb
-        size={AvatarSizes.pinnedConversation}
-        members={memberData}
-        style={{ marginLeft: 16, alignSelf: "center" }}
-      />
+      <PinnedV3GroupConversation group={conversation as GroupWithCodecsType} />
```
thierryskoda, line 19: Normally we shouldn't need the `as` if the type assertion is done correctly

components/PinnedConversations/PinnedV3DMConversation.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const preferredName = usePreferredName(peer);
+  const preferredAvatar = usePreferredAvatarUri(peer);
+
+  const onLongPress = usePinnedConversationLongPress(conversation.topic);
+  const onPress = useCallback(() => {
+    navigate("Conversation", { topic: conversation.topic });
+  }, [conversation.topic]);
+
+  const title = preferredName;
+  const showUnread = false;
```
thierryskoda, line 28: unnecessary IMO

components/PinnedConversations/PinnedV3GroupConversation.tsx
```diff
@@ -0,0 +1,66 @@
+import { PinnedConversation } from "./PinnedConversation";
```
thierryskoda, line 1: LOVE how the 2 (groups and DMS) are separated 🙏

components/PinnedConversations/PinnedV3DMConversation.tsx
alexrisch: Yeah mostly a setup for when that would exist, but just not finished work

components/PinnedConversations/PinnedV3Conversation.tsx
alexrisch: Yeah, I've made a PR for the SDK to resolve that: https://github.com/xmtp/xmtp-react-native/pull/525

PR #1170, opened 2024-11-07
feat: new eslint config

thierryskoda:
* Using ESlint v9
* Added some new rules
* Building custom rules is now easier
* The main file to look at is `eslint.config.mjs`


eslint.config.mjs
```diff
@@ -0,0 +1,144 @@
...
+  //   languageOptions: {
+  //     parser: parserTypescript,
+  //   },
+  //   plugins: {
+  //     "custom-plugin": customPlugin,
+  //   },
+  //   rules: {
+  //     "custom-plugin/single-function-per-file": "off", // Off for now
+  //   },
+  // },
```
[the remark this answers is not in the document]
thierryskoda: Totally agree that this rules makes no sense. I did it first to help with @alexrisch comments on having smaller utils files to make testability better but I agree with internal functions it might get chaotic. I'm also not expert in tests so will let @alexrisch maybe jump in here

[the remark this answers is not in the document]
alexrisch: Agree with @Koleok, testability issues is due to what's getting imported rather than what's getting exported

I'm personally all for this lint rule, but I believe I am likely in the minority, so I'm fine to not have it. I think I would also be the first person to start breaking it too 😆

custom-eslint-plugin/padding-before-react-hooks.js
```diff
@@ -0,0 +1,61 @@
+/** @type {import('eslint').Rule.RuleModule} */
```
alexrisch, line 1: Hmm who likes this rule 😄

features/onboarding/Privy/PrivyPhoneEntry.tsx
```diff
@@ -7,19 +7,20 @@ import {
   textSecondaryColor,
 } from "@styles/colors";
 import { spacing } from "@theme/spacing";
-import { AsYouType, CountryCode } from "libphonenumber-js";
+import type { CountryCode } from "libphonenumber-js";
+import { AsYouType } from "libphonenumber-js";
```
alexrisch, line 11: Just curious - Did these get split because of the new linter? Or something you caught yourself?

package.json
```diff
@@ -8,7 +8,8 @@
     "build": "node scripts/build/build.js",
     "commit": "git show --format=\"%h\" --no-patch",
     "db": "ts-node scripts/migrations/db.ts",
-    "lint": "eslint --ext .ts,.tsx . --max-warnings 420",
+    "lint": "eslint --max-warnings 2500",
```
alexrisch, line 11: Is this the current warnings?
I think we should be working towards reducing the lint warnings so I normally like to cap it at the current amount and start fixing and reducing as much as I can over time

polyfills.ts
```diff
@@ -1,20 +1,19 @@
...
 import "@walletconnect/react-native-compat";
 import "react-native-get-random-values";
-import "@ethersproject/shims";
 
-import "text-encoding";
 import "@azure/core-asynciterator-polyfill";
-import "react-native-polyfill-globals/auto";
 import { Crypto as WebCrypto } from "@peculiar/webcrypto";
 import "@stardazed/streams-polyfill";
-import ViewReactNativeStyleAttributes from "react-native/Libraries/Components/View/ReactNativeStyleAttributes";
```
alexrisch, line 11: Are you sure we don't need this any more?

polyfills.ts
```diff
@@ -1,20 +1,19 @@
 // Random values must be imported first for security.
+import "@ethersproject/shims";
```
alexrisch, line 2: Did you do a quick sanity run?
I think sometimes the polyfills require a certain ordering to work, so may be better to actually ignore any ordering in this particular file

package.json
thierryskoda: I think it's ~1300 but it's because I added some new rules.

I think during our refactoring we shouldn't care about that but I think the rules are important so it highlights the "wrong patterns"

polyfills.ts
thierryskoda: From what I saw this is not needed anymore since `scaleY` is already in `View` component

polyfills.ts
thierryskoda: No didn't do sanity run. I've put back the order and put a comment to not change the order just in case

Koleok: Second this 👍

alexrisch (approved):

features/onboarding/Privy/PrivyPhoneEntry.tsx
thierryskoda: Oh sorry never answered. ESLint was giving me an error for that

PR #1192, opened 2024-11-13
Allow ethereum deeplinks inside frames

nmalzieu:


components/Chat/Frame/FramePreview.tsx
```diff
@@ -25,6 +25,11 @@ import {
 } from "../../../utils/frames";
 import { MessageToDisplay } from "../Message/Message";
 
+const AUTHORIZED_URL_PROTOCOLS = [
```
alexrisch, line 28: Can we make this a set instead?
And move it out to it's own file

nmalzieu: @alexrisch done

alexrisch:
Rebased this to main branch to include in a new release: https://github.com/ephemeraHQ/converse-app/pull/1198

PR #1217, opened 2024-11-19
fix: Android V3 Build

alexrisch:
Fix Android Build
Update Conversation List to have hooks


thierryskoda (approved):
LGTM! Some hooks that we might review later once I merged my PR since it might be smart to reuse some stuff but we can do that later after my changes. I'll rebase when it's merged and take a deeper look

PR #1256, opened 2024-11-26
feat: messaging attachments + new dm convo improvements 

thierryskoda:


features/conversation/dm-conversation.screen.tsx
```diff
@@ -0,0 +1,281 @@
+/**
+ *
+ * WORK IN PROGRESS!
```
alexrisch, line 3: I think it makes sense to decouple new DM experience, I'm not 100% sure about decoupling DMs and Groups

I think there is value in accepting a topic and it "just working"
I think there is a lot of value to be had in accepting simpler links for devs to be able to make things on

alexrisch (approved):

PR #1258, opened 2024-11-26
feat: Notification Subscriptions forV3

alexrisch:
Added notification subscribing for V3 only

## Summary by CodeRabbit

- **New Features**
	- Enhanced notification subscription logic with improved error handling and logging.
	- Updated conversation fetching process for better synchronization.

- **Bug Fixes**
	- Improved error handling during synchronization to ensure responsiveness to connection issues.

- **Refactor**
	- Simplified notification handling and conversation tracking within the chat store.
	- Updated method signatures and types for clarity and maintainability.

- **Chores**
	- Removed legacy code and unnecessary subscriptions to streamline functionality.

### DIFF
--- a/utils/notifications.ts
+++ b/utils/notifications.ts
@@ -1,14 +1,20 @@
 import * as Notifications from "expo-notifications";
-import { debounce } from "perfect-debounce";
 import { Platform } from "react-native";
 
 import { ConversationId, ConversationTopic } from "@xmtp/react-native-sdk";
 import { currentAccount, useAccountsStore } from "../data/store/accountsStore";
 import { useAppStore } from "../data/store/appStore";
-import api from "./api";
+import api, { saveNotificationsSubscribe } from "./api";
 import { getTopicFromV3Id } from "./groupUtils/groupId";
 import mmkv from "./mmkv";
 import { navigate, navigateToTopic } from "./navigation";
+import { getXmtpClient } from "./xmtpRN/sync";
+import {
+  ConversationWithCodecsType,
+  ConverseXmtpClientType,
+} from "./xmtpRN/client";
+import logger from "./logger";
+import { savePushToken } from "./keychain/helpers";
 
 let nativePushToken: string | null;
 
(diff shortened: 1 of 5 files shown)


utils/notifications.ts
```diff
@@ -40,161 +46,98 @@ export const lastNotifSubscribeByAccount: {
...
+  try {
+    logger.info(
+      "[subscribeToNotifications] start",
+      account,
+      conversations.length
+    );
+    const thirtyDayPeriodsSinceEpoch = Math.floor(
+      Date.now() / 1000 / 60 / 60 / 24 / 30
+    );
+    const client = (await getXmtpClient(account)) as ConverseXmtpClientType;
```
thierryskoda, line 67: I see we do A LOT of those `const client = (await getXmtpClient(account)) as ConverseXmtpClientType;` could we make it simpler and at least not need the `as ... ` just curious

utils/notifications.ts
```diff
@@ -40,161 +46,98 @@ export const lastNotifSubscribeByAccount: {
...
+    };
+
+    logger.info("[subscribeToNotifications] saving notifications subscribe");
+    await saveNotificationsSubscribe(
+      account,
+      nativePushToken,
+      nativeTokenQuery.type,
+      nativeTokenQuery.type === "android" ? "converse-notifications" : null,
+      topicsToUpdateForPeriod
+    );
```
thierryskoda, line 133: Many params , would love a object instead 👀

utils/xmtpRN/sync.ts
```diff
@@ -146,8 +146,8 @@ const syncClientConversationList = async (account: string) => {
     });
     // Prefetch the conversation list so when we land on the conversation list
     // we have it ready, this will include syncing all groups
-    const conversationList = await prefetchConversationListQuery(account);
-    // TODO: Update notification topics
+    const conversationList = await fetchConversationListQuery(account);
+    subscribeToNotifications({ conversations: conversationList, account });
   } catch (e) {
     logger.error(e, {
       context: `Failed to fetch persisted conversation list for ${account}`,
```
thierryskoda, line 153: would be cool to start handling errors better. Maybe worth waiting until we have a clear strategy tho

utils/notifications.ts
alexrisch: Yeah I agree we don't need it, it's there because of old web stuff, so we should probably remove it

Issue #222, opened 2024-07-02
Searching for a user should return groups they are in with you


saulmc:

alexrisch:
Added this in V3 Release, closing this out

PR #1261, opened 2024-11-26
feat: V3 Search

alexrisch:
Added search for V3 Only

Added tests
Updated Snapshots

Closes https://github.com/ephemeraHQ/converse-app/issues/1262 
Closes https://github.com/ephemeraHQ/converse-app/issues/222

## Summary by CodeRabbit

- **New Features**
	- Enhanced search functionality for Direct Messages (DMs) and groups based on user-defined queries.
	- Introduced filtering mechanism in the Conversation List to improve search results based on conversation type.

- **Bug Fixes**
	- Corrected naming inconsistencies in inbox profile socials queries for improved clarity.

- **Tests**
	- Added unit tests for query matching functions to ensure accurate search results.

### DIFF
--- a/features/conversations/utils/search.ts
+++ b/features/conversations/utils/search.ts
@@ -83,9 +83,6 @@ const inboxIdMatchesSearchQuery = async ({
   searchQuery,
   inboxId,
 }: InboxIdSearchParams): Promise<boolean> => {
-  if (!inboxId) {
-    return false;
-  }
   if (inboxId.toLowerCase().includes(searchQuery.toLowerCase())) {
     return true;
   }
--- a/features/conversations/utils/search.ts
+++ b/features/conversations/utils/search.ts
@@ -1,0 +1,122 @@
+import { getInboxProfileSocialsQueryData } from "@/queries/useInboxProfileSocialsQuery";
+import { getPreferredInboxName } from "@/utils/profile";
+import type {
+  DmWithCodecsType,
+  GroupWithCodecsType,
+} from "@/utils/xmtpRN/client";
+import type { InboxId } from "@xmtp/react-native-sdk";
+
+type DmSearchParams = {
+  account: string;
+  searchQuery: string;
+  dm: DmWithCodecsType;
+};
+
+export const dmMatchesSearchQuery = async ({
+  account,
+  searchQuery,
+  dm,
+}: DmSearchParams): Promise<boolean> => {
+  if (!dm) {
+    return false;
+  }
+  const inboxId = await dm.peerInboxId();
+  if (await inboxIdMatchesSearchQuery({ account, searchQuery, inboxId })) {
+    return true;
+  }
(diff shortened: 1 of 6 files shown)


utils/xmtpRN/sync.ts
```diff
@@ -117,7 +117,7 @@ const syncClientConversationList = async (account: string) => {
     // Load the persisted conversation list
     await fetchPersistedConversationListQuery(account);
     // Streaming conversations
-    retryWithBackoff({
+    await retryWithBackoff({
```
thierryskoda, line 120: Something we could do in this one is not use `await` but have a .catch for each one. Since we don't want that if one has a bug to crash the others, but still want to catch the error.

Or do a `Promise.allSettled` maybe to have them all run at same time

features/conversations/utils/search.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+  return false;
+};
+
+type AddressSearchParams = {
+  searchQuery: string;
+  address: string;
+};
+
+const addressMatchesSearchQuery = async ({
```
thierryskoda, line 114: Why is this async?

features/conversations/utils/search.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+const inboxIdMatchesSearchQuery = async ({
+  account,
+  searchQuery,
+  inboxId,
+}: InboxIdSearchParams): Promise<boolean> => {
+  if (!inboxId) {
+    return false;
+  }
```
thierryskoda, line 94: Why do we have this if in the typing inboxId cannot be null or undefined

features/conversations/utils/search.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export const groupMatchesSearchQuery = async ({
+  account,
+  searchQuery,
+  group,
+}: GroupSearchParams): Promise<boolean> => {
+  if (!group) {
+    return false;
+  }
```
thierryskoda, line 54: why this? if the type says it can't be null or undefined

features/conversations/utils/search.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export const dmMatchesSearchQuery = async ({
+  account,
+  searchQuery,
+  dm,
+}: DmSearchParams): Promise<boolean> => {
+  if (!dm) {
+    return false;
+  }
```
thierryskoda, line 22: Same as others

features/conversations/utils/search.ts
alexrisch: You don't really lose anything from checking this and types are not final, but will remove

utils/xmtpRN/sync.ts
alexrisch: Removed for now, it was just there when I was trying to track down a streaming bug

PR #1280, opened 2024-12-03
fix: Swipable too sensitive

alexrisch:
Added check for translation to handle read/unread setting

### DIFF
--- a/components/V3DMListItem.tsx
+++ b/components/V3DMListItem.tsx
@@ -152,8 +152,11 @@ export const V3DMListItem = ({ conversation }: V3DMListItemProps) => {
   }, [topic]);
 
   const onLeftSwipe = useCallback(() => {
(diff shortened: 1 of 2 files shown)


components/V3DMListItem.tsx
```diff
@@ -152,8 +152,11 @@ export const V3DMListItem = ({ conversation }: V3DMListItemProps) => {
   }, [topic]);
 
   const onLeftSwipe = useCallback(() => {
-    toggleReadStatus();
-  }, [toggleReadStatus]);
+    const translation = ref.current?.state.rowTranslation;
+    if (translation && (translation as any)._value > 100) {
```
alexrisch, line 156: Would love to not do this, but moving stuff over from V2

thierryskoda (approved):

PR #1288, opened 2024-12-03
Ml/nobranch add dependency control pattern

technoplato:


jest.config.ts
technoplato, line 1: this was a fix for environment variables not being properly hydrated during tests

alexrisch (approved):

PR #1313, opened 2024-12-05
fix: Update List Group Metadata

alexrisch:
Update List Group metadata queries when needed
Fixes https://github.com/ephemeraHQ/converse-app/issues/1309 

## Release Notes

- **New Features**
  - Streamlined handling of group metadata updates including name, image, and description through new utility functions.
  - Added functions for updating group metadata in conversation lists, enhancing overall management.

- **Improvements**
  - Enhanced conversation list management with new functions for updating group properties.
  - Improved type safety in conversation list queries.

- **Bug Fixes**
  - Addressed case sensitivity issues in conversation list keys.

### DIFF
--- a/utils/groupUtils/handleGroupNameUpdate.ts
+++ b/utils/groupUtils/handleGroupNameUpdate.ts
@@ -1,0 +1,18 @@
+import { setGroupNameQueryData } from "@/queries/useGroupNameQuery";
+import { updateGroupNameToConversationListQuery } from "@/queries/useV3ConversationListQuery";
+import type { ConversationTopic } from "@xmtp/react-native-sdk";
+
+type HandleGroupNameUpdateParams = {
+  account: string;
+  topic: ConversationTopic;
+  name: string;
+};
+
+export const handleGroupNameUpdate = ({
+  account,
+  topic,
+  name,
+}: HandleGroupNameUpdateParams) => {
+  setGroupNameQueryData(account, topic, name);
+  updateGroupNameToConversationListQuery({ account, topic, name });
+};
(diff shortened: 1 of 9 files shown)


utils/groupUtils/handleGroupNameUpdate.ts
```diff
@@ -0,0 +1,18 @@
...
+import { updateGroupNameToConversationListQuery } from "@/queries/useV3ConversationListQuery";
+import type { ConversationTopic } from "@xmtp/react-native-sdk";
+
+type HandleGroupNameUpdateParams = {
+  account: string;
+  topic: ConversationTopic;
+  name: string;
+};
+
+export const handleGroupNameUpdate = ({
```
technoplato, line 11: it seems like rather than making helpers to perform two updates to two different data structures/queries, it would be simpler to consume the source of truth query in a derived query - is that not possible?

alexrisch: Which would be the source of truth in this case?

One is an item within a list of objects
The other is for the detail view

The list cannot use the methods from the detail view as that would have Number of rendered list items * 4 requests across the React Native bridge and connections on the Database in LibXmtp, which probably becomes too much to handle appropriately

The detail would require fetching the entire list, which is probably not what we want especially if we are landing into a conversation from a deep ink

There is work that would reduce the amount of connections to the DB, and we are getting close to being able to enable new architecture, maybe at that point the list data is driving the information, but I don't believe it's doable at the moment

technoplato: Ah, I think the context about heavy handed db lookups helped me bridge my gap of misunderstanding. Still getting my head around our data model

technoplato (approved):

utils/groupUtils/handleGroupNameUpdate.ts
alexrisch: I think there is definitely value in having a source of truth, but i think it's a bit difficult if we want the SDK to be our source of truth but also have some memory logic we need to do etc.

It's by no means perfect at the current moment

technoplato: I think we could greatly benefit from looking at using `react-query` more idiomatically for this, as it is exactly what it is designed for.

PR #1320, opened 2024-12-06
fix: Persisted queries

alexrisch:
Added persistance by query
Bumped ReactQuery Versions
Added handling to fetch persisted conversations before dismissing the splash screen

Closes https://github.com/ephemeraHQ/converse-app/issues/1321

## Release Notes

- **New Features**
  - Enhanced hydration logic for fetching conversation lists, improving asynchronous handling and logging.
  - Added persistence for query data in inbox and profile socials queries.

- **Improvements**
  - Refactored query configurations for conversation lists to streamline setup and improve maintainability.

- **Dependency Updates**
  - Updated several `@tanstack` library packages to version `5.62.2` for improved performance and stability.


components/StateHandlers/HydrationStateHandler.tsx
```diff
@@ -18,11 +19,29 @@ export default function HydrationStateHandler() {
...
         getInstalledWallets(false);
       }
-      accounts.map((a) => getXmtpClient(a));
 
-      // accounts.map((address) => {
-      //   refreshProfileForAddress(address, address);
-      // });
+      // Fetching persisted conversation lists for all accounts
+      // We may want to fetch only the selected account's conversation list
+      // in the future, but this is simple for now, and want to get feedback to really confirm
```
technoplato, line 25: What are we confirming?

queries/useV3ConversationListQuery.ts
technoplato, line 1: Just a nit on naming “new” things. We should just call all new things as they would be (conversationList) and rename any old stuff to v2 or deprecated or whatever in order to avoid a rename or any confusion moving forward.

queries/useInboxProfileSocialsQuery.ts
```diff
@@ -48,6 +49,7 @@ const inboxProfileSocialsQueryConfig = (
   // And automatic retries if there was an error fetching
   refetchOnMount: false,
   staleTime: 1000 * 60 * 60 * 24,
+  persister: reactQueryPersister,
```
technoplato, line 52: What is your reasoning for adding persistence just to these two queries?

alexrisch: The are plain JS objects that don't have methods on them

The classes that are returned from the SDK are a bit more involved to persist at the current moment.

It's added on the top level query client, but that doesn't seem to be working, not sure if it is a bug of react query or what

queries/useV3ConversationListQuery.ts
alexrisch: Yep, as Thierry called out would like to make naming  and folder structure changes later on when things have calmed down a bit in the repo

components/StateHandlers/HydrationStateHandler.tsx
alexrisch: Confirming actual timing of real users

Difficult to gauge in dev, but benchmarks on the SDK side are good

queries/useInboxProfileSocialsQuery.ts
technoplato: 10 4 - still find it super ugh we can't just get a plain JSON message, conversation, etc but we can definitely massage that on our side.

I think that's where we should aim for: 

pluck off plain json data (including calling async methods such as members, etc)
store that plain json data in cache
perform plucking in fetch function

See: https://tkdodo.eu/blog/react-query-data-transformations
CleanShot 2024-12-06 at 15 46 00@2x

Instead of mapping our todos to uppercase, we can pluck out all of the values we want from that sdk call.

PR #1333, opened 2024-12-10
fix: Update Conversation List when new conversation is streamed

alexrisch:
Add new conversation to list when streamed
Fix content type errors

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for fetching spam scores and conversation updates.
	- Added functions for updating group image, name, and description in conversation lists.
	- Introduced a new method for adding conversations to the conversation list.

- **Bug Fixes**
	- Improved robustness of content type handling to prevent errors with undefined values.

- **Documentation**
	- Updated method signatures for clarity on parameter types.

### DIFF
--- a/utils/xmtpRN/contentTypes/index.ts
+++ b/utils/xmtpRN/contentTypes/index.ts
@@ -26,7 +26,10 @@ export function isContentType(args: {
   return contentType.startsWith(prefix);
 }
 
-export function getMessageContentType(contentType: string) {
+export function getMessageContentType(contentType: string | undefined) {
+  if (!contentType) {
+    return undefined;
+  }
   return ObjectTyped.keys(contentTypesPrefixes).find((key) =>
     contentType.startsWith(contentTypesPrefixes[key])
   );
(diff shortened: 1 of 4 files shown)


thierryskoda (approved):

utils/xmtpRN/contentTypes/index.ts
```diff
@@ -26,7 +26,10 @@ export function isContentType(args: {
   return contentType.startsWith(prefix);
 }
 
-export function getMessageContentType(contentType: string) {
+export function getMessageContentType(contentType: string | undefined) {
+  if (!contentType) {
+    return undefined;
+  }
```
thierryskoda, line 32: I wonder if we should put | undefined here 🤔 

And instead handle the undefined before calling the function.

Curious on your thoughts. It's just that I feel it's counterintuitive to call this function with a contentType that is undefined haha.

Love the `| undefined` instead of `?: string` tho! At least it shows it wants an argument

thierryskoda (approved):

PR #1337, opened 2024-12-10
feat: Members in Group Title

alexrisch:
Added Group Members Count to group header title
Found additional translations needed and updated them Added French translations
Simulator Screenshot - iPhone 15 Pro - 2024-12-09 at 23 53 11


lourou (approved):
I like those french translations 🥐

PR #1386, opened 2024-12-19
Group invite deep link fix

lourou:
Deep linking was causing crashes due to attempts to access the `routes` property in an undefined navigation state. This occurred because the navigation state wasn't properly initialized before processing URLs, and also, handling for such `inviteId` was missing.

### Testing

Please verify/double check deep linking functionality for:
- Group invites
- DM links
- Regular group links
- Invalid/malformed links

Fixes [#1380](https://github.com/ephemeraHQ/converse-app/issues/1380)

### DIFF
--- a/features/GroupInvites/joinGroup/JoinGroupNavigation.tsx
+++ b/features/GroupInvites/joinGroup/JoinGroupNavigation.tsx
@@ -10,6 +10,7 @@ import {
 import { Platform, useColorScheme } from "react-native";
 
 import { JoinGroupScreen } from "./JoinGroup.screen";
+import { stackGroupScreenOptions } from "@/screens/Navigation/navHelpers";
 
 export type JoinGroupNavigationParams = {
   groupInviteId: string;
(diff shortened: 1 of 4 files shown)


features/GroupInvites/joinGroup/JoinGroupNavigation.tsx
```diff
@@ -21,20 +22,23 @@ export const JoinGroupScreenConfig = {
...
-        headerTitle: "",
-        headerTintColor:
-          Platform.OS === "android"
-            ? textSecondaryColor(colorScheme)
-            : textPrimaryColor(colorScheme),
-        animation: navigationAnimation,
-        headerTitleStyle: headerTitleStyle(colorScheme),
-      })}
-    />
+    <NativeStack.Group screenOptions={stackGroupScreenOptions(colorScheme)}>
```
thierryskoda, line 27: Not sure why we need a Group here 🤔 

But okay for now since we'll need to do a refactor on the navigation anyway

thierryskoda (approved):
LGTM! But @technoplato should check this also since he worked on that

screens/Navigation/navHelpers.ts
```diff
@@ -56,6 +56,28 @@ export const getConverseStateFromPath =
     if (pathForState?.startsWith("/")) {
       pathForState = pathForState.slice(1);
     }
+
+    // Handle group invite links
+    if (pathForState?.startsWith("group-invite")) {
```
alexrisch, line 61: I don't think we should need this any more, we should just match whatever the backend is giving us with the app

features/GroupInvites/joinGroup/joinGroup.machine.ts
```diff
@@ -233,18 +233,27 @@ export const joinGroupMachineLogic = setup({
 
       const invitedConversation: ConversationWithCodecsType | undefined =
         conversation.byId[invitedConversationId];
+
+      if (!invitedConversation) {
```
technoplato, line 237: this and the below logic seem more reasonable to me! Thanks @lourou

technoplato (approved):

screens/Navigation/navHelpers.ts
lourou: @alexrisch happy to remove this if we're only using Universal Links. Looks like the case at the moment – [see here on the website repo](https://github.com/ephemeraHQ/converse-website/blob/93f1aa5949e1f1231dc4b33086d3fe5bbe2a870a/utils/device.ts#L48-L52) but there is still a commented-out `${scheme}://` for group links specifically – I'm unsure; what do you think?

Is support for the URL app scheme no longer needed?

I did the change below on the website codebase in dev only (not committed) to use app scheme and use the group invite link, as Universal Links were not working locally:
```
: inviteId
-      ? `${url}/group-invite/${inviteId}`
+      ? `${scheme}://group-invite?inviteId=${inviteId}`
```

screens/Navigation/navHelpers.ts
```diff
@@ -56,6 +56,28 @@ export const getConverseStateFromPath =
...
+      const url = new URL(`https://${config.websiteDomain}/${pathForState}`);
+      const params = new URLSearchParams(url.search);
+      const inviteId = params.get("inviteId") || url.pathname.split("/")[1];
```
alexrisch, line 64: Sorry I meant this part, we shouldn't have to change the parameters to make it work here, we should just match what the backend is sending within our linking definition/screen parameters

alexrisch (approved):

PR #1395, opened 2024-12-19
Ml/fix deeplink to groupchat

technoplato:
https://www.loom.com/share/d1967df857fb4d1984cbdce195ff7739

## Summary by CodeRabbit

- **New Features**
	- Introduced a utility function to manage app store hydration.
	- Added navigation reference management to enhance navigation handling.

- **Bug Fixes**
	- Improved error handling and logging for navigation actions.

- **Refactor**
	- Streamlined conversation handling and navigation logic for better performance and clarity.
	- Simplified state management in the navigation component.
	- Replaced custom sleep function with a standardized wait function for delays in group joining processes.

- **Chores**
	- Updated import statements for cleaner code.


data/store/appStore.ts
```diff
@@ -85,3 +86,29 @@ export const useAppStore = create<AppStoreType>()(
...
+ */
+export const waitForXmtpClientHydration = async () => {
+  logger.debug(
+    `[waitForAppStoreHydration] Waiting for app store hydration to complete`
+  );
+  while (!useAppStore.getState().hydrationDone) {
+    logger.debug(
+      `[waitForAppStoreHydration] App store hydration not complete, waiting 100ms...`
+    );
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
alexrisch, line 111: Why are you doing this vs creating a subscription to the store instead?

thierryskoda: Yes I thought about that too. But It would look something like that no? Which isn't cleaner IMO or even much more performant. Maybe we keep what Michael have for now which is good enough IMO.

And we'll probably have to refactor stuff anyway later


```
const waitForReady = (): Promise<void> => {
  return new Promise((resolve) => {
    const { isReady } = useAccountsStore.getState();
    if (isReady) {
      resolve();
      return;
    }
    const unsubscribe = useAccountsStore.subscribe(
      (state) => state.isReady,
      (ready) => {
        if (ready) {
          resolve();
          unsubscribe(); // Stop listening after `isReady` becomes true
        }
      }
    );
  });
};
```

technoplato: you guys are right I like this better. It would also resolve sooner (rather than waiting for the 100ms window to elapse)

technoplato: here's a loom with this change: https://www.loom.com/share/d1967df857fb4d1984cbdce195ff7739 (explaining the reason for the wait(1)

features/GroupInvites/joinGroup/JoinGroup.client.ts
```diff
@@ -24,11 +24,11 @@ import { JoinGroupResult } from "./joinGroup.types";
 import { ConversationListQueryData } from "@/queries/useConversationListQuery";
 import { entify } from "@/queries/entify";
 import { GroupWithCodecsType } from "@/utils/xmtpRN/client";
+import { wait } from "@/utils/wait";
```
technoplato, line 27: this change is because I realized we had a utility for this already

features/notifications/utils/onInteractWithNotification.ts
```diff
@@ -58,40 +55,13 @@ export const onInteractWithNotification = async (
     | undefined;
 
   if (conversationTopic) {
-    // todo(lustig): zod verification of external payloads such as those from
-    // notifications, deep links, etc
+    await waitForXmtpClientHydration();
```
technoplato, line 58: Turns out all the crap I was doing before in the previous diff uh was done in client hydration so it's not needed here

screens/Main.tsx
```diff
@@ -84,9 +85,10 @@ export default function Main() {
           theme={navigationTheme}
           linking={linking}
           ref={(r) => {
-            if (r) {
-              converseNavigations["main"] = r;
-            }
+            logger.info(
+              `[Main] Setting navigation ref to ${r ? "not null" : "null"}`
+            );
+            setConverseNavigatorRef(r);
```
technoplato, line 91: We only have one navigation ref now that we are no longer supporting different layouts and stuff so I did some refactoring to reflect that

screens/Main.tsx
```diff
@@ -105,9 +107,7 @@ export default function Main() {
 const NavigationContent = () => {
   const authStatus = useAuthStatus();
 
-  const { splashScreenHidden } = useAppStore(
-    useSelect(["notificationsPermissionStatus", "splashScreenHidden"])
-  );
+  const { splashScreenHidden } = useAppStore(useSelect(["splashScreenHidden"]));
```
technoplato, line 110: unused

thierryskoda (approved):

PR #1404, opened 2024-12-19
fix: Fixed type to match new SDK Version

alexrisch:
Added senderInboxId

## Summary by CodeRabbit

- **New Features**
	- Implemented a polling mechanism for group join requests, improving the handling and status reporting of join attempts.
- **Bug Fixes**
	- Updated the property name for sender identification in message handling, enhancing clarity in the messaging process.


alexrisch:
> does this fix `.send` is not a function?

Hmm no I don't think so, I'll try to find a log for it, likely related to conversation context not getting set correctly

thierryskoda:
> does this fix `.send` is not a function?

should have been fixed in previous PR. You've seen it again?

Issue #1359, opened 2024-12-15
Account Refactor


alexrisch:
# Proposed Refactor Plan: XMTP Inbox IDs

This document outlines a plan to refactor our messaging app so that each user’s “inbox” is keyed by an **XMTP inbox ID** instead of an Ethereum address. The goal is to support various wallet types (embedded, smart-contract, external) and also introduce incognito (ephemeral) accounts without doing a massive rewrite. Below is a staged plan, justification, some high-level architecture details, and an FAQ.

---

## 1. Overview

Currently, we have the concept of `account = Ethereum address`. This is becoming cumbersome because:

- We want to support passkey-based _embedded wallets_ (via Turnkey, etc.).
- We eventually want _smart contract wallets_ for feeless transactions, advanced approvals, or social recovery.
- We also have incognito/ephemeral accounts that don’t require passkeys or wallet top-ups.

Rather than continuing to treat an Ethereum address as the primary key everywhere, we plan to move to an **XMTP inbox ID** model. That single ID then links to any wallets (embedded, smart-contract, external), user profile data (ENS, handle, photo), and advanced settings (incognito mode, ephemeral key material, etc.).

**Key links and references**

https://github.com/xmtp/XIPs/blob/main/XIPs/xip-46-multi-wallet-identity.md

- [Figma: Converse App Flows](https://www.figma.com/design/p6mt4tEDltI4mypD3TIgUk/Converse-App)
- [Notion: Smart Contract Wallet Work](https://www.notion.so/ephemerahq/Smart-Contract-Wallet-Work-Refinement-11330823ce9280539056d14fcc4e0de7?pvs=4)
- [Notion: Approvals in Converse](https://www.notion.so/ephemerahq/Approvals-in-Converse-0ca3419d992c4acb853654b808a338ff?pvs=4)
- [Excalidraw Diagram (Read-Only)](https://link.excalidraw.com/readonly/WFBJrBWRAz9IOU8COzZi)
- Meetings
  - [Michael and Alex Planning](https://www.loom.com/share/cd7f9576f29449e1a7e4f6e7ad515364)

---

### **Stage 1: Move from `account` to `inbox ID`**

1. **Replace all references**
   - Everywhere in the front end that references an “account” as an Ethereum address should become `inboxId`.
   - Create a simple map for `inboxId -> ethereumAddress` for cases where we need to instantiate the XMTP client, since XMTP currently needs an address to be instantiated.
2. **Update / break old data**
[...]

PR #1488, opened 2025-01-08
Account => Inbox ID Gigantic Refactor

technoplato:
Looking to get some early feedback on this big boy

https://github.com/ephemeraHQ/converse-app/issues/1359 phase 1


thierryskoda (commented):
* I think we can totally kill Privy. @alexrisch maybe can confirm but if it's causing you problem in your refractor I would just kill everything that has Privy in it
* I would remove the explicit return type like `: Promise<ConversationDataEntity | undefined>`. I know lots of them aren't you that put it, but just in case you see some I think you can remove while doing refactor.
* Maybe out of scope in this PR but we'll need to refactor a lot of hooks and queries to stop passing the `currentInboxId` since we'll always execute based on the currentInboxId by default
* I would maybe change your cursor rule to not add the big comments. Since we're not doing that pattern it will be weird to see it just from time to time. And they are so big that I will never read IMO.
* I think account.utils will be SOO big. So I would already start to try to find a way to maybe create folder and separate things? I'm not sure about putting `getXmptClient` there for example.
* Love replacing the `address` with `ethAddress`. Feels more explicit. 
* Like we discussed I would always favor the function having the typing it needs and checking in the caller if we have the value. Example in `useGroupQuery`, it shouldn't be `inboxId: string | undefined;` but `inboxId: InboxId` and we'll do `inboxId!` in the caller if necessary and have our "safe" hook etc...

technoplato:
gonna close them for now it'll always be in github if we want to reference it but our api cleanup will doa lot of good here

PR #1533, opened 2025-01-16
feat: add react query expo devtools (shift+m in console to use)

technoplato:
https://www.loom.com/share/cf5db6a3d0234996938c4ae0d30c5c15

yarn start
shift m
choose react-query devtools
explore


thierryskoda (approved):
Nice! Never used it but it's true that it could be helpful

PR #1534, opened 2025-01-16
fix: Optimistic sending message + Improved react-query cache + Many many other fixes

thierryskoda:


features/consent/update-addresses-consent-for-account.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  logger.debug(
+    `[XMTPRN Contacts] Consenting to addresses on protocol: ${addresses.join(
+      ", "
+    )}`
+  );
+  const start = new Date().getTime();
+
+  if (consent === "allow") {
```
technoplato, line 29: nit: just derivce state from consent and perform the loop once with no conditionals

features/consent/update-addresses-consent-for-account.ts
technoplato, line 1: this name/function is confusing to me: am I allowing/denying accounts from messaging me?

if so id name update-consent-for-address-for-current-account

a little wordy but captures meaning better imo

features/consent/update-groups-consent-for-account.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (!client) {
+    throw new Error("Client not found");
+  }
+
+  logger.debug(
+    `[XMTPRN Contacts] Consenting to groups on protocol: ${groupIds.join(", ")}`
+  );
+  const start = new Date().getTime();
+
+  const state = consent === "allow" ? "allowed" : "denied";
```
technoplato, line 27: do this in the for users one but we also probably want to throw if invalid consent, but that'd be a typescript issue anyway so you're choice whether you want to extract this to a function or not

features/consent/use-allow-group.mutation.ts
```diff
@@ -121,7 +119,7 @@ export const getAllowGroupMutationOptions = (
         group.topic,
         context.previousConsent || "unknown"
       );
-      updateConversationInConversationsQuery({
+      updateConversationInConversationsQueryData({
```
technoplato, line 122: metadata or data?

features/conversation-list/conversation-list-item/conversation-list-item-dm.tsx
```diff
@@ -48,11 +53,11 @@ export const ConversationListItemDm = memo(function ConversationListItemDm({
   }, [avatarUri, preferredName, theme]);
 
   const onPress = useCallback(() => {
-    prefetchConversationMessages(currentAccount, topic);
+    prefetchConversationMessages(currentAccount, conversationTopic);
```
technoplato, line 56: not as part of this pr, but we should have messages for conversations loaded earlier than this...

features/conversation-list/conversation-list-item/conversation-list-item-dm.tsx
```diff
@@ -73,15 +78,17 @@ export const ConversationListItemDm = memo(function ConversationListItemDm({
...
+      return (
+        <ToggleUnreadSwipeableAction {...args} topic={conversationTopic} />
+      );
     },
-    [topic]
+    [conversationTopic]
   );
 
-  const deleteDm = useDeleteDm(conversation);
+  const deleteDm = useDeleteDm(conversation as DmWithCodecsType);
```
technoplato, line 88: again, not as part of this but we should start deriving these hooks from more primitive values (ids for example) as a best practice

features/conversation-list/conversation-list.screen.tsx
```diff
@@ -48,12 +43,24 @@ export function ConversationListScreen(props: IConversationListProps) {
...
 
   useHeaderWrapper();
 
+  // For now, let's make sure we always are up to date with the conversations
+  // useScreenFocusEffectOnce(() => {
+  //   refetchConversations().catch(captureError);
+  // });
+
+  // For now, let's make sure we always are up to date with the conversations
+  useAppStateHandlers({
```
technoplato, line 58: shouldn't we just use query options to refetch on foreground?

features/conversation-list/hooks/use-delete-dm.ts
```diff
@@ -42,26 +42,23 @@ export const useDeleteDm = (dm: DmWithCodecsType) => {
...
         account: currentAccount,
         topic: dm.topic,
-        context: "useDeleteDm",
       })?.isDeleted;
 
-      updateConversationDataQueryData({
+      updateConversationMetadataQueryData({
         account: currentAccount,
         topic: dm.topic,
-        context: "useDeleteDm",
```
technoplato, line 54: I thoubght you added this context pattern @thierryskoda why are you removing now?

features/conversation-list/hooks/useMessagePlainText.ts
```diff
@@ -7,37 +7,53 @@ import {
...
+  useEffect(() => {
+    if (!message) {
+      return;
+    }
+
     if (isGroupUpdatedMessage(message)) {
-      const idsToFetch: InboxId[] = [];
-      const content = message.content();
-      idsToFetch.push(content.initiatedByInboxId);
+      // TODO: check why we need to cast here. We should not need to do this since we have type guards
```
technoplato, line 39: is there an is/as in isGroupUpdatedMesssage

features/conversation-list/use-conversation-list-conversations.tsx
```diff
@@ -0,0 +1,45 @@
+import { useCurrentAccount } from "@/data/store/accountsStore";
+import { isConversationAllowed } from "@/features/conversation/utils/is-conversation-allowed";
+import { getConversationMetadataQueryOptions } from "@/queries/conversation-metadata-query";
+import { getConversationsQueryOptions } from "@/queries/conversations-query";
+import { useQueries, useQuery } from "@tanstack/react-query";
+import { useMemo } from "react";
+
+export const useConversationListConversations = () => {
```
technoplato, line 8: does `useConversationsList` work here?

technoplato: or maybe `useConversationsMetadataForConversationsList`????? names... amirite

features/conversation-list/use-conversation-list-conversations.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const conversationsDataQueries = useQueries({
+    queries: (conversations ?? []).map((conversation) =>
+      getConversationMetadataQueryOptions({
+        account: currentAccount!,
+        topic: conversation.topic,
+      })
+    ),
+  });
+
+  const conversationsFiltered = useMemo(() => {
```
technoplato, line 26: I believe we can use `combine` for this: 



Motivation for doing so is using an array in the dependency array generally blows up in my experience due to reference vs value comparison

features/conversation/conversation-preview/conversation-preview-messages.query.ts
```diff
@@ -0,0 +1,33 @@
...
+import logger from "@/utils/logger";
+import { useQuery, UseQueryOptions } from "@tanstack/react-query";
+import { ConversationTopic } from "@xmtp/react-native-sdk";
+
+export const useConversationPreviewMessages = (
+  account: string,
+  topic: ConversationTopic,
+  options?: Partial<UseQueryOptions<ConversationMessagesQueryData>>
+) => {
+  return useQuery({
```
technoplato, line 15: I think as best practice we always extgract into const getConversationPreviewMessageQueryOPtions = ...

features/conversation/conversation.tsx
```diff
@@ -296,7 +295,15 @@ const Messages = memo(function Messages(props: {
             isLatestMessageSentByCurrentUser={
               latestMessageIdByCurrentUser === messageId
             }
-            animateEntering={index === 0}
+            animateEntering={
+              index === 0 &&
+              // Need this because otherwise because our optimistic updates, we first create a dummy message with a random id
```
technoplato, line 300: we shouldn't create the optimistic message with random id but hash the conversationId, userid and message content and maybe tag it with the prefix `optimistic:` or some other way to deliniate between the two.

queries/use-conversation-messages-query.ts
```diff
@@ -88,21 +80,21 @@ export function refetchConversationMessages(
...
 ) {
-  logger.info("[refetchConversationMessages] refetching messages");
+  logger.debug("[refetchConversationMessages] refetching messages");
   return queryClient.refetchQueries(
     getConversationMessagesQueryOptions(account, topic)
   );
 }
 
-export const addConversationMessage = (args: {
+export const addConversationMessageQuery = (args: {
```
technoplato, line 89: I don't think this should be suffixed query, but we should have a way to deliniate direct mutations to the cache vs updates out in the real world. 

maybe optimisticAddConversationMessage?

features/consent/use-allow-group.mutation.ts
thierryskoda: data. It's just how i always name those function. To say that I'm updating the data of the query

features/notifications/utils/background/notificationContent.ts
```diff
@@ -1,68 +1,76 @@
...
     if (action === "added" && schema === "unicode") {
-      // Checking if the group message is from me
-      const isFromMe = await isGroupMessageFromMe(
-        group.client,
-        referencedMessageId
-      );
-      if (!isFromMe) return;
+      // If the message is from the current account, we don't need to show the notification
+      // TODO: This check was there before but not sure if it's needed because we still want to show the notification
+      //  if someone else reacted to a message from the current account?
```
technoplato, line 53: great callout i just wanted to prevent 

i send message
i see notification for my message

in my understanding, a reaction would technically be a message from another user - please let me knmow where i'm wrong

ios/ConverseNotificationExtension/Xmtp/Client.swift
```diff
@@ -35,10 +35,10 @@ func getXmtpClient(account: String) async -> XMTP.Client? {
     let groupId = "group.\(try! getInfoPlistValue(key: "AppBundleId", defaultValue: nil))"
     let groupDir = (FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: groupId)?.path)!
     let client = try await Client.build(address: account, options: .init(api: .init(env: xmtpEnv), dbEncryptionKey: encryptionKey, dbDirectory: groupDir))
-    client.register(codec: AttachmentCodec())
-    client.register(codec: RemoteAttachmentCodec())
-    client.register(codec: ReactionCodec())
-    client.register(codec: ReplyCodec())
+    Client.register(codec: AttachmentCodec())
```
technoplato, line 38: is this global for all clients?

queries/conversation-metadata-query.ts
```diff
@@ -0,0 +1,104 @@
...
+async function getConversationMetadata(args: IArgs) {
+  logger.debug(
+    `[ConversationMetadataQuery] getConversationMetadata for account: ${args.account}, topic: ${args.topic}`
+  );
+  return batchedGetConversationMetadata.fetch(args);
+}
+
+const batchedGetConversationMetadata = create({
+  scheduler: windowScheduler(50),
+  resolver: (items, query) => {
```
technoplato, line 68: items is vague here is there a better word for this arg?

queries/useConversationMessages.ts
```diff
@@ -25,47 +21,43 @@ export type ConversationMessagesQueryData = Awaited<
   ReturnType<typeof conversationMessagesQueryFn>
 >;
 
-export const conversationMessagesQueryFn = async (
-  conversation: ConversationWithCodecsType,
```
technoplato, line 29: nice! I'm doing the exact same thing.

queries/use-conversation-messages-query.ts
```diff
@@ -25,47 +21,43 @@ export type ConversationMessagesQueryData = Awaited<
...
 >;
 
-export const conversationMessagesQueryFn = async (
-  conversation: ConversationWithCodecsType,
-  options?: MessagesOptions
-) => {
-  const start = performance.now();
-  logger.info("[useConversationMessages] queryFn fetching messages...");
+export const conversationMessagesQueryFn = async (args: {
+  account: string;
```
technoplato, line 25: love we're both aligned here - I was just doing this

utils/api/auth.ts
technoplato, line 1: love the breakup!

utils/update-object-and-methods.ts
```diff
@@ -0,0 +1,12 @@
+export function updateObjectAndMethods<T>(
```
technoplato, line 1: pthis needs documentation because its super unclear its required for optimistic updates etc to function properly

technoplato (approved):
holy hell that's a large one but im good merging and applying feedback as next pr

features/consent/use-allow-group.mutation.ts
technoplato: perfect just clarifying

technoplato: perfect just clarifying

PR #1554, opened 2025-01-18
fix: Many fixes

thierryskoda:


queries/useConversationQuery.ts
```diff
@@ -13,15 +15,44 @@ type IArgs = {
...
+
+  // Try to find conversation in local DB first
+  let conversation = await client.conversations.findConversationByTopic(topic);
+
+  // If not found locally, sync and try again
+  if (!conversation) {
+    logger.warn(
+      `[useConversationQuery] Conversation not found in local DB, syncing conversations`
+    );
+    await client.conversations.sync();
```
alexrisch, line 37: Is this syncing all conversations now?
If not it will cause the list to sync, but conversations will be out of date

queries/unknown-consent-conversations-query.ts
```diff
@@ -7,9 +7,12 @@ import { UseQueryOptions } from "@tanstack/react-query";
...
   return (
-    conversations?.filter((conversation) =>
-      // Only the unknown conversations
-      isConversationConsentUnknown(conversation)
+    conversations?.filter(
+      (conversation) =>
+        // Only the unknown conversations
+        isConversationConsentUnknown(conversation) &&
+        // Only the conversations with last message
+        conversation.lastMessage
```
alexrisch, line 15: I don't think we should be checking on lastMessage here, there's times where it can be empty

We will need to also make sure the requests isn't incrementing based on those either

thierryskoda: Ahh I thought conversations without lastMessage would always be non-relevant but good will remove.

features/conversation-list/hooks/use-delete-dm.ts
```diff
@@ -7,49 +7,54 @@ import {
...
 
-export const useDeleteDm = (dm: DmWithCodecsType) => {
+export const useDeleteDm = ({ topic }: { topic: ConversationTopic }) => {
   const { theme } = useAppTheme();
   const colorScheme = theme.isDark ? "dark" : "light";
 
   const currentAccount = useCurrentAccount()!;
 
+  const { data: conversationId } = useQuery({
+    ...getConversationQueryOptions({
```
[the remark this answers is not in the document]
thierryskoda: IMO that doesn't change anything since it's with a useQuery but might be wrong

features/conversation-list/hooks/use-conversation-list-item-context-menu-default-props.tsx
```diff
@@ -138,18 +140,29 @@ function useConversationContextMenuDeleteItem(args: {
...
 
-  const handleDeleteFn = conversation
-    ? isConversationGroup(conversation)
-      ? // eslint-disable-next-line react-hooks/rules-of-hooks
-        useDeleteGroup({ groupTopic: conversationTopic })
-      : // eslint-disable-next-line react-hooks/rules-of-hooks
-        useDeleteDm(conversation)
-    : () => null;
+  const handleDelete = useCallback(async () => {
+    if (!conversation) return;
```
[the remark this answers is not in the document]
thierryskoda: 200%!

features/consent/use-dm-consent.tsx
```diff
@@ -16,17 +16,17 @@ import {
...
 import { updateInboxIdsConsentForAccount } from "./update-inbox-ids-consent-for-account";
 
-export function useDmConsent(args: {
-  peerInboxId: InboxId;
-  conversationId: ConversationId;
-  topic: ConversationTopic;
-}) {
-  const { peerInboxId, conversationId, topic } = args;
-
+export function useDmConsent() {
```
technoplato, line 19: ForCurrentInbox/Account

features/consent/use-dm-consent.tsx
```diff
@@ -16,17 +16,17 @@ import {
...
   const currentAccount = useCurrentAccount()!;
 
   return useMutation({
-    mutationFn: async (args: { consent: "allow" | "deny" }) => {
+    mutationFn: async (args: {
```
technoplato, line 23: does moving the arguments to the functions instead of the hook fix anything or is it just more idiomatic?

thierryskoda: I just feel like if we're only using the args when calling the mutation, might as well pass them when we call the mutation.

alexrisch (approved):

PR #1555, opened 2025-01-19
feat: New deployment flow

thierryskoda:


App.tsx
```diff
@@ -1,21 +1,21 @@
...
 import { queryClient } from "@queries/queryClient";
 import { MaterialDarkTheme, MaterialLightTheme } from "@styles/colors";
 import { QueryClientProvider } from "@tanstack/react-query";
-import { useReactQueryDevTools } from "@dev-plugins/react-query";
+import "expo-dev-client";
+import {
+  configureReanimatedLogger,
+  ReanimatedLogLevel,
+} from "react-native-reanimated";
+import "./polyfills";
```
alexrisch, line 16: This needs to be imported before most other libraries

scripts/build/increment-version.js
```diff
@@ -44,33 +45,37 @@ if (!incrementTypeArg) {
...
+    incrementType
+  );
 
-  appJson.expo.version = newVersion;
-  appJson.expo.ios.buildNumber = "1";
-  appJson.expo.android.versionCode = appJson.expo.android.versionCode + 1;
+  // Update Expo version and platform-specific version numbers
+  // iOS: Reset build number to 1 for each version update
+  // Android: Increment the version code by 1
+  versionsJson.expo.version = newAppVersion;
```
alexrisch, line 64: There's no more expo property in the file

.github/workflows/build-or-update-prod.yml
```diff
@@ -0,0 +1,80 @@
...
+          fetch-depth: 2
+
+      - name: Get Commit Message
+        id: commit_msg
+        run: echo "::set-output name=message::$(git log -1 --pretty=%B)"
+
+      - name: Check for dependency or config changes
+        id: check-changes
+        run: |
+          if git diff --name-only HEAD^ HEAD | grep -E "package.json|app.config.ts"; then
```
technoplato, line 40: What about pods? They shouldn't but could change even if these files don't

App.tsx
```diff
@@ -1,21 +1,21 @@
-import "expo-dev-client";
-import "./polyfills";
-import {
-  configureReanimatedLogger,
-  ReanimatedLogLevel,
-} from "react-native-reanimated";
 import { configure as configureCoinbase } from "@coinbase/wallet-mobile-sdk";
 import DebugButton from "@components/DebugButton";
 import { BottomSheetModalProvider } from "@design-system/BottomSheet/BottomSheetModalProvider";
+import { useReactQueryDevTools } from "@dev-plugins/react-query";
```
technoplato, line 4: What import cleanup rule do you have? I've used prettier import cleanup before - maybe we should automate and consolidate our different import sorting strategies

app.config.ts
```diff
@@ -1,79 +1,120 @@
 import { ConfigContext, ExpoConfig } from "expo/config";
-import warnOnce from "warn-once";
-// Removed this import as it was causing a build error
-// import type { PluginConfigTypeAndroid } from "expo-build-properties/src/pluginConfig";
+import appConfig from "./versions.json";
 
-import appBuildNumbers from "./app.json";
+const expoEnv = process.env.EXPO_ENV as Environment;
```
technoplato, line 4: Todo for later: zod parse and verify configuration shape and provide helpful descriptions of what everything is

environment.d.ts
```diff
@@ -0,0 +1,24 @@
+declare global {
+  namespace NodeJS {
+    interface ProcessEnv {
+      EXPO_ENV: string;
```
technoplato, line 4: Zod will be helpful here too to prevent needing to declare a type and then check at runtime

utils/api/api.config.ts
```diff
@@ -31,6 +31,10 @@ export const getApiUri = () => {
   // Get configured dev API URI
   const envApiUri = process.env.EXPO_PUBLIC_DEV_API_URI;
 
+  if (!envApiUri) {
```
technoplato, line 34: We should move this env validation to the very start of the application using zod to parse and derive types for env.d.ts

utils/xmtpRN/client/client.ts
```diff
@@ -0,0 +1,55 @@
...
+  new ReadReceiptCodec(),
+  new GroupUpdatedCodec(),
+  new ReplyCodec(),
+  new RemoteAttachmentCodec(),
+  new StaticAttachmentCodec(),
+  new TransactionReferenceCodec(),
+  new CoinbaseMessagingPaymentCodec(),
+];
+
+export const getXmtpClientFromAddress = async (address: string) => {
```
technoplato, line 31: We should name this build .... as it isn't just getting

technoplato (approved):

App.tsx
thierryskoda: Screenshot 2025-01-21 at 8 27 46 AM

utils/xmtpRN/client/client.ts
thierryskoda: Yes this whole client file need a refactor

PR #1567, opened 2025-01-22
fix: Requests list items

thierryskoda:


features/consent/use-allow-group.mutation.ts
```diff
@@ -126,6 +131,10 @@ export const getAllowGroupMutationOptions = (
           state: context.previousConsent,
         },
       });
+      addConversationToUnknownConsentConversationsQuery({
+        account,
+        conversation: group,
```
technoplato, line 136: do we want to be consistent and use topic (id) instead of object (group)

thierryskoda: you have to pass the whole object here because it's not even existing in cache yet

PR #1570, opened 2025-01-22
react-querify Search

technoplato:
It's now buttery smooth

https://www.loom.com/share/3543d66f12484709b071a30bebbcff81


features/create-conversation/components/MessageSection.tsx
```diff
@@ -1,9 +1,14 @@
 import React from "react";
```
thierryskoda, line 1: why lots of files don't have lowercase kebab? 🤔

features/create-conversation/components/MessageSection.tsx
```diff
@@ -12,20 +17,44 @@ import { MessageSectionProps } from "../create-conversation.types";
  * @param props.isError - Whether the message is an error
  */
 export function MessageSection({ message, isError }: MessageSectionProps) {
-  const { themed } = useAppTheme();
+  const { themed, theme } = useAppTheme();
+  const { profile } = useProfile();
+
+  const inviteThemStyle: TextStyle = {
+    fontSize: 14,
+    lineHeight: 18,
```
thierryskoda, line 25: we should get those with our theme

thierryskoda: or even better with textStyles

queries/search-query.ts
```diff
@@ -0,0 +1,167 @@
...
+  searchQuery: string;
+  addressesToOmit: string[];
+};
+
+type ISearchResult = {
+  message: string;
+  profileSearchResults: Record<string, IProfileSocials>;
+};
+
+import { currentAccount, getCurrentAccount } from "@/data/store/accountsStore";
```
thierryskoda, line 16: this shouldn't be in imports

queries/search-query.ts
```diff
@@ -0,0 +1,167 @@
...
+import { setProfileRecordSocialsQueryData } from "@/queries/useProfileSocialsQuery";
+import { isEmptyObject } from "@/utils/objects";
+import { shortAddress } from "@/utils/strings/shortAddress";
+import { accountCanMessagePeer } from "@/features/consent/account-can-message-peer";
+import logger from "@/utils/logger";
+
+/**
+ * Handles searching when searchQuery is a supported domain/ENS
+ */
+async function handlePeerSearch(searchQuery: string): Promise<ISearchResult> {
```
thierryskoda, line 29: `: Promise<ISearchResult>` :(

queries/search-query.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  if (isSupportedPeer(searchQuery)) {
+    return handlePeerSearch(searchQuery);
+  }
+
+  return handleGeneralSearch(searchQuery);
+}
+
+const SearchResultStaleTime = 1000 * 60 * 5; // 5 minutes
```
thierryskoda, line 129: Why extract as a constant?

queries/search-query.ts
```diff
@@ -0,0 +1,167 @@
...
+    currentAccountAddress,
+  ].map((address) => address.toLowerCase());
+
+  for (const address of Object.keys(profileSearchResults || {})) {
+    if (allAddressesToOmit.includes(address)) {
+      if (profileSearchResults) {
+        delete profileSearchResults[address];
+      }
+    }
+  }
```
thierryskoda, line 158: we can use the `select` in `useQuery` to do the filtering

thierryskoda (approved):

features/create-conversation/components/MessageSection.tsx
technoplato: Blame Claude

features/create-conversation/components/MessageSection.tsx
technoplato: Agree

queries/search-query.ts
technoplato: 🤣

queries/search-query.ts
technoplato: No magic numbers

queries/search-query.ts
technoplato: I don’t want to use that because it would prevent those users from entering the cache and this cache is an installation wide cache used by all users so doing it manually is correct behavior I believe

technoplato: ohhhh nvm i left that comment tired - you right

PR #1571, opened 2025-01-23
feat: Persistence

alexrisch:
Removed Account from inboxID query
Added react query persistence for conversations, profiles, members

### DIFF
--- a/features/GroupInvites/joinGroup/JoinGroup.client.ts
+++ b/features/GroupInvites/joinGroup/JoinGroup.client.ts
@@ -27,6 +27,7 @@ import { JoinGroupResult } from "./joinGroup.types";
 import { IConversationsQuery } from "@/queries/use-conversations-query";
 import { entify } from "@/queries/entify";
 import { wait } from "@/utils/general";
+import { InstallationId } from "@xmtp/react-native-sdk/build/lib/Client";
 
 const GROUP_JOIN_REQUEST_POLL_MAX_ATTEMPTS = 10;
 const GROUP_JOIN_REQUEST_POLL_INTERVAL_MS = 1000;
@@ -297,12 +298,12 @@ export class JoinGroupClient {
       const fixtureGroup: GroupData = {
         id: GroupIdUserAlreadyWasAMemberOf,
         createdAt: new Date().getTime(),
-        members: async () => [],
+        clientInstallationId: "clientInstallationId123" as InstallationId,
+        version: ConversationVersion.GROUP,
         topic: "topic123" as ConversationTopic,
         // has user been blocked?
         isGroupActive: true,
         state: "allowed",
-        creatorInboxId: async () => "0xabc" as InboxId,
         name: "Group Name",
         addedByInboxId: "0x123" as InboxId,
         imageUrlSquare: "https://www.google.com",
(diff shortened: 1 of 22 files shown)


features/conversation-list/hooks/useMessagePlainText.ts
```diff
@@ -49,14 +46,13 @@ export const useMessagePlainText = (
       });
       if (inboxIds.length > 0) {
         for (const inboxId of inboxIds) {
-          fetchInboxProfileSocialsQuery(account!, inboxId).catch(captureError);
+          fetchInboxProfileSocialsQuery(inboxId).catch(captureError);
```
technoplato, line 49: Can we rename this to fetchProfileSocialsBynboxId?

queries/useInboxProfileSocialsQuery.ts
```diff
@@ -70,44 +56,32 @@ const inboxProfileSocialsQueryConfig = (
   refetchOnMount: false,
   staleTime: 1000 * 60 * 60 * 24,
   initialData: (): IProfileSocials[] | null | undefined => {
```
technoplato, line 58: Could we remove this as it's handled by persistence now and will only be empty on first load?

queries/utils/persistence.ts
```diff
@@ -0,0 +1,61 @@
...
+import { DEFAULT_GC_TIME } from "../queryClient.constants";
+import mmkv from "@/utils/mmkv";
+import logger from "@/utils/logger";
+import { QueryState } from "@tanstack/query-core";
+
+export type GenericPersistedQuery<T, E = Error> = Omit<
+  PersistedQuery,
+  "state"
+> & {
+  state: QueryState<T | null, E>;
```
technoplato, line 15: Why are we retyping state here?

features/GroupInvites/joinGroup/JoinGroup.client.ts
```diff
@@ -464,17 +465,15 @@ export class JoinGroupClient {
       } as const;
     };
 
-    const fixtureFetchGroupsByAccount = async (
-      account: string
-    ): Promise<GroupsDataEntity> => {
+    const fixtureFetchGroupsByAccount = async (): Promise<GroupsDataEntity> => {
       const fixtureGroup: GroupData = {
+        clientInstallationId: "clientInstallationId123" as InstallationId,
```
nplasterer, line 470: You should be able to get this? A reason this is hardcoded?

queries/useInboxProfileSocialsQuery.ts
alexrisch: Ah yes good catch!

features/conversation-list/hooks/useMessagePlainText.ts
alexrisch: It matches more of what the backend is actually doing

There are Profiles and Inbox Profiles

Profiles is a 1 to 1 with an Ethereum address
Inbox Profile is several resolved Profiles from the Addresses of the Inbox Id that we then resolve into a displayed Profile

queries/utils/persistence.ts
alexrisch: It's more a helper, everything is unknown with the PersistedQuery so just helps to type it later if we have to do some custom deserializing

technoplato:
this has some great starts - will reference it once we want full persistnece,, closing for now cc @thierryskoda @lourou

PR #1581, opened 2025-01-29
feat: Profile styling enhancements

lourou:
This updates the Profile screen to consistently use surface background color throughout the screen **and** navigation header, replacing the default surfaceless background.

Changes:
- Updated Profile screen's container background to use surface color
- Added backgroundColor to useHeader configuration to ensure navigation header uses surface color
- Removed border bottom from sections to maintain visual consistency
- Improved visual hierarchy with border top between sections

Added documentation in `useHeader` explaining the header background convention:
- Headers are 'surfaceless' by default (for content screens like conversation list and chat)
- Settings-like screens (Profile, AppSettings) explicitly set surface background


features/profiles/profile.screen.tsx
```diff
@@ -130,6 +130,7 @@ export function ProfileScreen() {
   // Header configuration
   useHeader(
     {
+      backgroundColor: theme.colors.background.surface,
```
thierryskoda, line 133: Should we put this backgroundColor by default to our Header instead?

thierryskoda (approved):
Small suggestion to always centralize styling we know we'll need to apply in lots of places

features/profiles/profile.screen.tsx
lourou: @thierryskoda the header already has [a default `surfaceless` background](https://github.com/ephemeraHQ/converse-app/blob/lr/profile-styling-update/design-system/Header/Header.tsx#L49) used in conversation list and in a conversation.

Here, for the Profile, I put the Header background as `surface` to use the same background color as the screen content.

Adding a commit with comments! Looks good to you?

PR #1588, opened 2025-02-03
fix: Custom Context Menu solution for Android

thierryskoda:


technoplato:
oops - I thought this had approval as the merge button was green 
I'll test now and let you know if issues
working on android stuff now

lourou:
Yes that's weird, can we put back at least 1 approval before we merge? I'm not an admin so can't do

PR #1590, opened 2025-02-04
fix(AnimatedText): improve animation handling & fix Android font weights

lourou:
- Separate text animation and style handling
- Fix Android font weight string conversion
- Add proper type safety for styles
- Improve style merging logic

Tested with AnimatedTextCarousel & verified on Android

### DIFF
--- a/components/animated-text-carousel.tsx
+++ b/components/animated-text-carousel.tsx
@@ -92,13 +92,7 @@ export const AnimatedTextCarousel = memo(function AnimatedTextCarousel({
     <AnimatedCenter>
       <AnimatableText
         text={textAV}
-        style={[
-          textStyle,
-          useAnimatedStyle(() => ({
-            opacity: textAS.opacity,
-            textAlign: "center",
-          })),
-        ]}
+        style={[textStyle, textAS, { textAlign: "center" }]}
       />
     </AnimatedCenter>
   );
--- a/components/animated-text-carousel.tsx
+++ b/components/animated-text-carousel.tsx
@@ -1,7 +1,6 @@
 import { AnimatedCenter } from "@/design-system/Center";
 import { AnimatableText } from "@/design-system/Text/AnimatedText";
 import { ITextProps } from "@/design-system/Text/Text.props";
-import { debugBorder } from "@/utils/debug-style";
 import React, { memo, useEffect, useRef } from "react";
 import {
   cancelAnimation,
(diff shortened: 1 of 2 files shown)


components/animated-text-carousel.tsx
```diff
@@ -93,7 +92,13 @@ export const AnimatedTextCarousel = memo(function AnimatedTextCarousel({
     <AnimatedCenter>
       <AnimatableText
         text={textAV}
-        style={[textStyle, textAS, { textAlign: "center" }]}
+        style={[
+          textStyle,
+          useAnimatedStyle(() => ({
+            opacity: textAS.opacity,
+            textAlign: "center",
+          })),
```
thierryskoda, line 100: why this?

design-system/Text/AnimatedText.tsx
```diff
@@ -15,31 +21,50 @@ export type IAnimatedTextProps = AnimatedProps<ITextProps>;
...
+const AnimatedAnimateableText =
+  Animated.createAnimatedComponent(AnimateableText);
+
 // A specialized text component that efficiently updates text content using SharedValue
 // Optimized for frequent text changes without causing re-renders
 // Uses react-native-animateable-text under the hood for better performance
-export type IAnimatableTextProps = RNTextProps & {
+export type IAnimatableTextProps = Omit<RNTextProps, "style"> & {
   text: SharedValue<string>;
+  style?: StyleProp<TextStyle>;
```
thierryskoda, line 32: can we put small comment why we do this?

design-system/Text/AnimatedText.tsx
```diff
@@ -15,31 +21,50 @@ export type IAnimatedTextProps = AnimatedProps<ITextProps>;
...
+      typeof style === "object" &&
+      !Array.isArray(style) &&
+      "fontWeight" in style
+    ) {
+      baseStyle.fontWeight = String(
+        style.fontWeight
+      ) as TextStyle["fontWeight"];
+    } else if (style && typeof style === "object" && !Array.isArray(style)) {
+      Object.assign(baseStyle, style);
+    }
```
thierryskoda, line 55: can we add a small comment why this. Also do we know if it's necessary? Seems very complicated.

Maybe instead we just add fontSize: number, fontWeight: string (why string?), lineHeight: number in props. So that anyone using that componnets knows that we cannot just pass anystyle to it because of Android issue? So we explicity list the available props and format it should be passed?

thierryskoda (changes requested):

design-system/Text/AnimatedText.tsx
lourou: Yes, definitely! Adding this :)

components/animated-text-carousel.tsx
lourou: I've simplified this

design-system/Text/AnimatedText.tsx
```diff
@@ -15,31 +21,58 @@ export type IAnimatedTextProps = AnimatedProps<ITextProps>;
 
 export const AnimatedText = Animated.createAnimatedComponent(Text);
 
+const AnimatedAnimateableText =
+  Animated.createAnimatedComponent(AnimateableText);
```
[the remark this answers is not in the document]
lourou: Fixed! Tested on Android and app launches successfully ✅

PR #1594, opened 2025-02-06
fix: Android context menu position offset

### LINKED ISSUE #1591: Bug: On Android, bottom part of context menu on message starts at wrong location
### Describe the bug

_No response_

### Expected behavior

_No response_

### Steps to reproduce the bug

_No response_

lourou:
This fixes an issue where the context menu on Android devices appeared at the wrong vertical position (too high).

Changes:
- Added status bar height consideration in MessageContextMenuContainer
- Adjusted the menu position calculation to offset for Android's status bar properly
- Replaced theme.layout.screen.height with useWindowDimensions for more accurate screen measurements
- Removed unused debugBorder import

Quick note: Fixed text cutoff issue in the segment controller on Android by adjusting the height from 32px to 36px and adding proper line height. This ensures the text "You might know" and "Hidden requests" are fully visible. ℹ️ This will need a complete refactor and new design, so I did this quick update.

Fixes #1591

### DIFF
--- a/features/conversation/conversation-message/conversation-message-context-menu/conversation-message-context-menu-container.tsx
+++ b/features/conversation/conversation-message/conversation-message-context-menu/conversation-message-context-menu-container.tsx
@@ -2,6 +2,7 @@ import { AnimatedVStack, VStack } from "@/design-system/VStack";
 import { useConversationMessageContextMenuStyles } from "@/features/conversation/conversation-message/conversation-message-context-menu/conversation-message-context-menu.styles";
 import { useAppTheme } from "@theme/useAppTheme";
 import { memo, useEffect } from "react";
+import { Platform, StatusBar, useWindowDimensions } from "react-native";
 import {
   useAnimatedStyle,
   useSharedValue,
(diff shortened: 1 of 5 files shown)


features/conversation/conversation-message/conversation-message-context-menu/conversation-message-context-menu-container.tsx
```diff
@@ -85,16 +86,22 @@ export const MessageContextMenuContainer = memo(
...
+    // On Android, we need to account for the status bar height by moving the menu down
+    const statusBarHeight =
+      Platform.OS === "android" ? StatusBar.currentHeight || 0 : 0;
+
     const distanceFromTop =
       itemRectY -
       MESSAGE_CONTEXT_MENU_ABOVE_MESSAGE_REACTIONS_HEIGHT -
-      verticalSpaceBetweenSections;
+      verticalSpaceBetweenSections +
+      (Platform.OS === "android" ? statusBarHeight : 0);
```
thierryskoda, line 104: Shouldn't we use `safeAreaInsets` top  like we're doing at other places? I know if we do that it might fuck iOS but can we undersatnd why? I feel using StatusBar height isn't good here since StatusBar height isn't really stable from what I know.

thierryskoda: Also I thought the problem was that only the bottom part of the context view menu was off. Not the entire context menu. Might be wrong here. Can't reproduce on my device since the app crashes.

Issue #1498, opened 2025-01-10
[Screen] Conversation Composer


nplasterer:
On click of the new conversation button
<img width="589" alt="Screenshot 2025-01-09 at 7 35 57 PM" src="https://github.com/user-attachments/assets/8af264f1-5341-4aad-aa55-90fe36d71e4c" />

Take you to a composer to create a new conversations
https://www.figma.com/design/p6mt4tEDltI4mypD3TIgUk/Converse-App?node-id=5026-27441&t=Tso4OHIbfHTlG5bz-0

<img width="325" alt="Screenshot 2025-01-09 at 7 41 08 PM" src="https://github.com/user-attachments/assets/935e0632-cd0d-412c-a36c-5f459945f0c0" />

users list designs: https://www.figma.com/design/p6mt4tEDltI4mypD3TIgUk/Converse-App?node-id=5191-4200&t=KDRZMuK1xpiNBKG9-11

- [x] Ability to type in user names and select them from a menu to add them to a conversation
- [x] Ability to type in a wallet address to the list (if the user does not have a convos identity)
  - [x] Check to see if this wallet address is on the V3 network or not and show an error otherwise
- [x] Show new group in the top right corner if more than 1 identity is pre-populated if only one identity show new convo
- [x] If the address is only 1 person and a message is typed and sent and a dm already exists with that user drop on the existing dm screen and send the message
- [x] If the address is only 1 person create a dm and send the message
- [x] If multiple addresses are used create a group chat and send the message

nplasterer:
@covrter can you give me a few visuals:

- [ ] What it would look like to start typing a name and how the pop up to select a name would look?
- [ ] What it would look like if I have a wallet address mixed in with the convos identities?

nplasterer:
Replaces
![Image](https://github.com/user-attachments/assets/6784d373-db75-43ac-9166-aa55a886ac78)

technoplato:
@nplasterer are we okay to consider sender-switching out of scope for this ticket?

![CleanShot 2025-01-10 at 11 12 00@2x](https://github.com/user-attachments/assets/326739dd-92c6-4ecc-baf7-90d8299dd89c)

nplasterer:
Yes that is out of scope.

nplasterer:
Compiling a first pass bug list
### Must
- [ ] group creation errors are not surfaced it just looks like nothing happened when creation fails (adding self to group, bad keypackage etc..)
[...]

PR #1600, opened 2025-02-07
fix: Many fixes in conversation composer + Faster animation + Better message context menu

thierryskoda:
https://github.com/ephemeraHQ/converse-app/issues/1498#issuecomment-2638374328


technoplato (approved):
yolo ship it

PR #1604, opened 2025-02-12
fix: Onboarding refactor and fixes

thierryskoda:


technoplato:
let's close this for now and come bck after march deadline if we like

PR #1607, opened 2025-02-18
Authentication System Overhaul PR

technoplato:
## Overview
This PR implements a comprehensive JWT-based authentication system with secure token management and automatic refresh handling.

### 1. JWT Authentication Implementation
- Added new JWT query management system (`jwt.query.ts`)
- Implemented secure JWT persistence using encrypted storage
- Added automatic JWT refresh mechanism on 401 errors

### 2. Authentication Headers System
- Created new header interceptor system that handles two types of auth:
  1. Authentication routes (`/api/v1/users`, `/api/v1/authenticate`) using cryptographic proof
  2. All other routes using JWT Bearer token
- Moved authentication-related code to dedicated `features/authentication` directory

### 3. Error Handling & Logout
- Added `useLogoutOnJwtRefreshError` hook for automatic logout on JWT refresh failures
- Implemented in `App.tsx` at the root level
- Added proper error handling and logging throughout authentication flow

### 4. API Reorganization
- Moved user creation API from `utils/api/users` to `features/authentication/authentication.api.ts`
- Added proper Zod schema validation for API responses
- Consolidated authentication-related types and interfaces

### 5. Documentation
- Added comprehensive `authentication.readme.md` explaining:
  - Authentication flow with Firebase AppCheck, XMTP, and Ethereum wallet
  - API endpoint details and header requirements
  - Security considerations and implementation details

### 6. Configuration Updates
- Added `reactQueryEncryptionKey` to config for secure JWT storage
- Updated environment variable handling for secure query encryption

### 7. Minor Cleanup
- Removed redundant account parameter from various API calls
- Updated import paths to reflect new organization
- Fixed logger imports to use consistent pattern


features/authentication/authentication.api.ts
```diff
@@ -0,0 +1,71 @@
...
+      privyAddress: smartContractWalletAddress,
+      xmtpId: inboxId,
+    },
+  };
+
+  const response = await api.post<CreateUserResponse>(
+    "/api/v1/users",
+    requestData
+  );
+  return createUserResponseSchema.parse(response.data);
```
thierryskoda, line 55: Wonder if we should do `safeParse` here. Just because let's say the Backend accitdently change 1 property and so this request no longer work but in reality it would work for 99% of the time.

features/authentication/authentication.api.ts
```diff
@@ -0,0 +1,71 @@
...
+    .nullable(),
+});
+
+type CreateUserResponse = z.infer<typeof createUserResponseSchema>;
+
+export const createUser = async (args: {
+  privyUserId: string;
+  smartContractWalletAddress: string;
+  inboxId: string;
+}): Promise<CreateUserResponse> => {
```
thierryskoda, line 36: Shouldn't be needed

utils/mmkv.ts
```diff
@@ -51,23 +52,48 @@ export const clearSecureMmkvForAccount = async (account: string) => {
...
+}
+
+export const reactQueryPersister = experimental_createPersister({
+  storage: createMMKVStorage(reactQueryMMKV),
   maxAge: DEFAULT_GC_TIME,
 });
+
+export const secureQueryPersister = experimental_createPersister({
+  storage: createMMKVStorage(secureQueryMMKV),
+});
```
thierryskoda, line 99: Is this really "secured"? Since the encryption key is our EXPO_PUBLIC var? 

I understand we need to try but in this case maybe we should use `expo-secure-store` which is more secure than `mmkv`.

Maybe we store the encryption key in expo-secure-store? Here's what ChatGPT gave me quickly:

````
import { MMKV } from 'react-native-mmkv';
import * as SecureStore from 'expo-secure-store'; // Secure storage

const STORAGE_KEY = 'mmkv_encryption_key';

// Function to get or generate a secure encryption key
async function getEncryptionKey() {
  let key = await SecureStore.getItemAsync(STORAGE_KEY);
  if (!key) {
    key = generateRandomKey(); // Generate a new key if none exists
    await SecureStore.setItemAsync(STORAGE_KEY, key);
  }
  return key;
}

// Function to generate a random encryption key
function generateRandomKey() {
  return [...Array(16)]
    .map(() => Math.floor(Math.random() * 256))
    .map((n) => String.fromCharCode(n))
    .join('');
}

// Initialize MMKV with encryption
async function setupMMKV() {
  const encryptionKey = await getEncryptionKey();
  const storage = new MMKV({
    id: 'encrypted-storage',
    encryptionKey: encryptionKey,
  });

  return storage;
}

setupMMKV().then((storage) => {
  storage.set('userToken', 'super-secret-token');
  console.log(storage.getString('userToken')); // Will return decrypted data
});
````

**Review by thierryskoda — APPROVED** [review]
Nice!! Looks clean!! 👌

File: features/authentication/authentication.api.ts (hunk 1 above)
Comment (rygine) [reply features/authentication/authentication.api.ts:55]: why are we doing client side validation of the response data at all? we own the backend and know exactly what it's going to return. seems like a lot of extra work when we really just need a type assertion.

File: features/authentication/authentication.readme.md [hunk 4]
```diff
@@ -0,0 +1,89 @@
...
+### Create User Endpoint
+
+**Purpose**: Creates a new user account
+
+**Headers** (all four required):
+
+1. **Installation ID**: Identifier for the XMTP inbox, derived from the device
+2. **Firebase AppCheck Token**: Provided by Firebase
+3. **Signed AppCheck Token**: The Firebase AppCheck token signed by the XMTP inbox (client-side)
+4. **Ethereum Wallet**: The wallet used for signing, obtained via the Privy passkey flow
```
rygine, line 54: when did this change?

features/authentication/authentication.api.ts
technoplato: just as a sanity check. not much extra work to define the type and parse with zod, so we just do it. 

even though we own it, it can still change and we want to always catch errors as early as possible, which would be at the repsonse site rather than downstream

technoplato: I think we just want to parse everything and keep it up to date. no sense in letting these get out of date. 

later @rygine had a really good idea about just building an sdk for the backend, so not going to bike shed too much ont hisatm

features/authentication/authentication.readme.md
technoplato: good catch - i'd mistranscribed to chatgpt

PR #1611, opened 2025-02-18
Ml/profile optimistic updates and api changes

technoplato:


features/current-user/create-user.query.ts
```diff
@@ -0,0 +1,139 @@
...
+  privyUserId: string;
+  smartContractWalletAddress: string;
+  inboxId: string;
+  profile: {
+    name: string;
+    avatar?: string;
+  };
+};
+
+const userQueryKey = () => ["user"];
```
thierryskoda, line 37: createUserQueryKey? and ["current-user"] ?

features/current-user/create-user.query.ts
```diff
@@ -0,0 +1,139 @@
...
+    onSuccess: (createdUser, variables, { optimisticProfile }) => {
+      setCurrentUserData(createdUser);
+      setProfileQueryData({
+        xmtpId: variables.inboxId,
+        data: optimisticProfile,
+      });
+      invalidateCurrentUserQuery();
+      invalidateProfileQuery({ xmtpId: variables.inboxId });
+    },
+  });
```
thierryskoda, line 138: might need an onError to remove queryData and even signOut the user?

features/current-user/create-user.query.ts
```diff
@@ -0,0 +1,139 @@
...
+      };
+    },
+    onSuccess: (createdUser, variables, { optimisticProfile }) => {
+      setCurrentUserData(createdUser);
+      setProfileQueryData({
+        xmtpId: variables.inboxId,
+        data: optimisticProfile,
+      });
+      invalidateCurrentUserQuery();
+      invalidateProfileQuery({ xmtpId: variables.inboxId });
```
thierryskoda, line 136: why invalidate? why not just:

setCurrentUserData(createdUser);
      setProfileQueryData({
        xmtpId: variables.inboxId,
        data: createdUser.profile,
      });

features/profiles/profiles.api.ts
```diff
@@ -12,86 +11,78 @@ const ProfileSchema = z.object({
...
 };
 
-export const getAllProfilesForUser = async ({
+export const fetchAllProfilesForUser = async ({
   convosUserId,
 }: {
   convosUserId: string;
-}): Promise<IProfile[]> => {
+}): Promise<IConvosProfileForInbox[]> => {
   const { data } = await api.get(`/api/v1/profiles/user/${convosUserId}`);
```
thierryskoda, line 43: Usually the best practices route would be more like `/api/v1/user/${convosUserId}/profiles`

thierryskoda (approved):
Couple small things but I think if you're confident with this we should merge since it touches things that I'm using in my PR

features/current-user/create-user.query.ts
technoplato: good call

features/profiles/profiles.api.ts
technoplato: just gonna follow what's there for now we can change later

features/current-user/create-user.query.ts
technoplato: i don't thiiink the user would be considered logged in here but adding just in case

